PR #39, opened 2020-10-16
swap trade functionality

creed-victor:


creed-victor:
changes available with #47

PR #66, opened 2020-11-23
Add subresource integrity check

tammaroivan:


creed-victor (approved):

creed-victor:
Also, if resource allows it, add checks to resources in `/public/index.html` file (tradingview, motomo tracker etc)

tammaroivan:
AFAIK is not possible as this generates the sha256/384 when building. If we want to do this we should generate a `sha `each time matomo/tradingview/etc. get's an update (because the sha will be different to the one we will have in the html) and it does not really makes sense.

creed-victor:
That's true. Will research if there is some tracker that may provide some kind of integrity.
There was an incident with other crypto project where attackers hacked analytics code, injected custom web3 and was able to manipulate transaction requests of their customers.

PR #71, opened 2020-11-26
Fix/active loans sort

BetsyBraddock:
I refactored the active loans table, and the sorting now works properly for all columns


src/app/components/ActiveUserLoanContainer/components/ActiveLoanTableMobile/index.tsx
```diff
@@ -24,33 +26,68 @@ interface Props {
...
+            {formatAsBTC(item.positionSize, item.currency)}
+          </div>
+          <div
+            className={`col-3 ${item.profit > 0 ? 'text-green' : 'text-red'}`}
+          >
+            ${' '}
+            {item.profit.toLocaleString('en', {
+              maximumFractionDigits: 4,
+              minimumFractionDigits: 4,
+            })}
```
creed-victor, line 73: For currency it's better to use toLocaleString with currency type:
```
{item.profit.toLocaleString('en', {
    style: 'currency',
    currency: 'USD',
    maximumFractionDigits: 4,
    minimumFractionDigits: 4,
})}
```
Instead of specifying locale `'en'` we can also set it to `undefined` - app would use users preferred locale in this case.

Also, this code is repeated for couple of times - maybe it's a good chance to create new helper function.

src/app/containers/TradingHistory/index.tsx
```diff
@@ -245,6 +245,7 @@ function HistoryTable(props: { items: CalculatedEvent[] }) {
 
   const data = React.useMemo(() => {
     return props.items.map(item => {
+      console.log(item);
```
creed-victor, line 248:
```suggestion
```

creed-victor (changes requested):
Remove console.log :)

src/app/components/ActiveUserLoanContainer/components/ActiveLoanTableMobile/index.tsx
BetsyBraddock: I've added in some helper functions, thanks! For currencies, I set the locale as 'en-US', because otherwise the price looks like 'US$1000' instead of '$1000' which I thought looked a bit crowded.

creed-victor (approved):

src/app/components/ActiveUserLoanContainer/components/ActiveLoanTableMobile/index.tsx
creed-victor: Thanks 💯

PR #78, opened 2020-12-01
create component with locale selector which change app's language

pietro-maximoff:
1. create component with locale selector which change app's language [en, es]
2. english texts move to json file


src/app/components/LanguageToggle/index.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from 'react';
+import i18next from 'i18next';
+
+export function LanguageToggle() {
+  const [currentLang, setCurrentLang] = useState(
+    i18next.language || window.localStorage.i18nextLng,
```
creed-victor, line 6: `window.localStorage` most likely will fail on safari browser if user browses incognito.
Use [reactjs-localstorage](https://www.npmjs.com/package/reactjs-localstorage) package (I think it's already in our dependencies) which make sure that app doesn't crash.

creed-victor (changes requested):

creed-victor (approved):

src/app/components/LanguageToggle/index.tsx
pietro-maximoff: done it

PR #101, opened 2020-12-10
Feat/email notification

BetsyBraddock:
Updated the email notification form for backend AWS SES mail server, and fixed the form so that it resets when dialog box is closed.

Also a few small style fixes to the email button and form.


creed-victor (changes requested):
I think we also need `REACT_APP_MAIL_SRV` in .env.testnet and .env.production files?

.env.production
```diff
@@ -10,4 +10,4 @@ REACT_APP_WHITELIST_TOKEN=0x576aE218aeCfD4CbD2DBe07250b47e26060932B1
 
 REACT_APP_YBUG_ID=3f1jrxvzrhkn1b975t8b
 
-REACT_APP_MAIL_SRV=https://mailservice.sovryn.app/
+REACT_APP_MAIL_SRV=http://3.13.8.41:3010/
```
creed-victor, line 13: Oh, noticed only now that url uses http and not https protocol.
Browser will decline requests to this endpoint when project is not in localhost.
Which server is that, I can enable https there if needed

creed-victor (changes requested):

creed-victor (approved):

PR #103, opened 2020-12-11
Feat/connect wallet tutorial

BetsyBraddock:
This is a temporary component for mobile, until we have figured out the best way for users to use the dapp from a mobile browser. It lets users register their email to hear when the mobile version is working, and connects to our existing Mailchimp mailing list.

It looks like: 
image


creed-victor (changes requested):
Was FastBTCTutorial changes from fast-btc branch you provided for me previuosly?
I moved your stylings to old fast-btc form and deleting FastBTCTutorial form, so if there wasn't any changes you should delete it too (otherwise user will see 2 dialogs)

BetsyBraddock:
> Was FastBTCTutorial changes from fast-btc branch you provided for me previuosly?
> I moved your stylings to old fast-btc form and deleting FastBTCTutorial form, so if there wasn't any changes you should delete it too (otherwise user will see 2 dialogs)

Thanks, I've deleted the FastBTCTutorial dialog

creed-victor (approved):

PR #116, opened 2020-12-15
Fixes #92 lend/borrow tabs for pt_br on mobile screen

### LINKED ISSUE #92: Align buttons (Emprestar) and (Tomar Empréstimo) side by side
On PT-BR language, align (Emprestar) and (Tomar Empréstimo) buttons side by side

photo_2020-12-05_23-22-59

acarnagey:
Updates html lang attribute when updating language in the footer so that the styling can know whether it should style a little bit differently based on the language it is in.  It makes the font size a little bit smaller so that the tabs fit in one row since there is more text for that language.

### DIFF
--- a/src/app/components/LanguageToggle/index.tsx
+++ b/src/app/components/LanguageToggle/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 
 import i18next from 'i18next';
 import { reactLocalStorage } from 'reactjs-localstorage';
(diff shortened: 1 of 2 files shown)


creed-victor (changes requested):
Thanks for collaborating, but I'm thinking that it would be better to add media query for mobile browsers instead of sticking font size to locale. In future we may have more languages that will broke it again.

Also, I think it's a good idea to update all tabs (not only for lending/borrow page), tabs shouldn't wrap and always be in row too, so maybe having them with `text-overflow: ellipsis;` also would help.

acarnagey:
@creed-victor The main thing that is causing the tab to go on the next row is the default rule `.nav { flex-wrap: wrap }` we could do nav no wrap and center vertically which would look like
Screen Shot 2020-12-16 at 12 18 12 PM
Another option could be nav no wrap and ellipsis non active tab(s)
nowrap-ellipsis-word
We could also make the active tab stretch across it's content width like this.
Screen Shot 2020-12-16 at 12 13 00 PM
I think these options would make sense for any screen size and that we wouldn't need to apply this to `@include media-breakpoint-down(xs)`
Let me know which option you are leaning towards.

creed-victor:
@acarnagey yep, 3rd option looks good for me :)

creed-victor (approved):

PR #121, opened 2020-12-17
Fix/active loan table

BetsyBraddock:
Fixed profit calculation and added another check to if popup tutorial should show (if user is already on RSK network)


creed-victor:
Math for short is still off:

Screenshot 2020-12-17 at 12 49 14

creed-victor (approved):

PR #190, opened 2021-01-15
Metamask discouragement modal

creed-victor:
Modal which will be forced upon user if they have MetaMask wallet extension installed until they accept it.
Modals purpose is to inform user that right now MetaMask is not the best choice for RSK chains, inform what can go wrong if they choose to use with our dapp and encourage use Nifty Wallet instead.

PR is draft and modal has only checkbox and confirmation button, still needs some work on design and adding copies.


kayalight:
[Uploading warning.xd.zip…]()

PR #254, opened 2021-01-22
[salesPage] - update styles and text  for BTC

pietro-maximoff:


src/app/containers/SendBTC/index.tsx
```diff
@@ -336,22 +336,22 @@ export default function SendBTC({ setShowCalc }) {
...
-                    Send BTC to receive SOV in your engaged wallet
+                    Send BTC to pre-order SOV in your engaged wallet
                   </li>
                   <li className="mb-2">
                     Do not send anything other than BTC to this address
                     otherwise your assets will be lost permanently
                   </li>
                   <li className="mb-2">
-                    Please allow up to 15 mins for the transaction to process
+                    Please allow up to xx mins for the transaction to process
```
creed-victor, line 346: `xx` minutes?

pietro-maximoff: mins in design

creed-victor: I mean't why it's xx, I think it should be number instead.
And because it's bitcoin transaction, then 15-60 minutes is very possible wait time.

pietro-maximoff: will ask Tzu now

creed-victor: I think you can use this

```suggestion
                    Please allow 5 to 60 mins for the transaction to process
```

PR #301, opened 2021-01-27
[Header] - fix mobile menu (fullscreen)

pietro-maximoff:


src/app/components/Header/index.tsx
```diff
@@ -62,18 +144,15 @@ export function Header() {
     );
   });
```
creed-victor, line 146: Add this to prevent user scrolling while menu is opened

```suggestion
React.useEffect(() => {
  const body = document.body;
  if (open) {
    body.classList.add('overflow-hidden');
  } else {
    body.classList.remove('overflow-hidden');
  }
   return () =>{
    body.classList.remove('overflow-hidden');
  }
}, [open])
```

creed-victor (changes requested):

Issue #143, opened 2021-01-12
I don't see strings to translate some texts to PT-BR


bitcoinuser:
Fixed

bitcoinuser:
![Captura de tela de 2021-01-12 14-17-26](https://user-images.githubusercontent.com/1178856/104349048-32b48c00-54e1-11eb-8bd0-e655e1eb015d.png)

bitcoinuser:
One more:

![Captura de tela de 2021-01-17 14-22-12](https://user-images.githubusercontent.com/1178856/104850754-ba2e4080-58cf-11eb-97c5-1a0003bcbb62.png)

bitcoinuser:
One more

![Captura de tela de 2021-01-17 15-30-57](https://user-images.githubusercontent.com/1178856/104852344-527cf300-58d9-11eb-92a4-8c3b64803170.png)

creed-victor:
"Liquidate position" and "Repay loan" dialog strings added.

bitcoinuser:
More:

![Captura de tela de 2021-01-22 22-12-57](https://user-images.githubusercontent.com/1178856/105564113-7b4d2000-5cff-11eb-89a5-a5e883a94150.png)

bitcoinuser:
More:

![photo_2021-01-23_21-26-50](https://user-images.githubusercontent.com/1178856/105618032-dc3d2c80-5dc1-11eb-8f74-8e0ad2c6a3c8.jpg)

bitcoinuser:
More

![Captura de tela de 2021-01-24 11-22-42](https://user-images.githubusercontent.com/1178856/105633379-f06f4100-5e36-11eb-8567-39f4f4ef280a.png)

bitcoinuser:
@soulBit Could add this ones too? Thanks

![Captura de tela de 2021-02-01 10-15-41](https://user-images.githubusercontent.com/1178856/106479377-08734000-6489-11eb-8978-a41a819043a9.png)

![Captura de tela de 2021-02-01 12-24-37](https://user-images.githubusercontent.com/1178856/106479404-0f9a4e00-6489-11eb-9c4e-3224c8066b35.png)

bitcoinuser:
Hi @soulBit, please add this, thanks

![Captura de tela de 2021-02-02 13-02-59](https://user-images.githubusercontent.com/1178856/106627624-bea55b00-6557-11eb-9587-7806d80fa70f.png)

PR #317, opened 2021-01-30
Add translation strings for #143

soulBit:
Added strings for all text that is referenced in #143, with the exception of this one (as I believe this is just a temporary warning):
https://github.com/DistributedCollective/Sovryn-frontend/issues/143#issuecomment-766356682

All of these will need appropriate translations, I have just copied over the english text.

There is one string that I spotted in NotificationFormContainer::86 that looks like it needs translation, but I'm not sure how you would like to handle this one as there are some newline characters and piping that probably shouldn't go into the translation json:
 `${timestamp} \n \n Please confirm that the details associated with this account will now be: \n \n Username: ${name} \n Email: ${email}`

I also reviewed the rest of the Components/Containers in the app and have found a number of other locations where translations should be applied - let me know if you'd like a list of those submitted as separate ticket/pull request.

### DIFF
--- a/src/app/components/MetaMaskDiscouragementNotifyModal/index.tsx
+++ b/src/app/components/MetaMaskDiscouragementNotifyModal/index.tsx
@@ -4,7 +4,10 @@
  *
  */
 import React, { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { translations } from '../../../locales/i18n';
 import { Checkbox } from '@blueprintjs/core';
+import styled from 'styled-components/macro';
 import { local } from '../../../utils/storage';
 import { Dialog } from '../../containers/Dialog/Loadable';
 import logo from './logo.svg';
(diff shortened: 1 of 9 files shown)


creed-victor:
Hey, thanks.
Yes thats temporary so it's no need to translate. 

`${timestamp} \n \n Please confirm that the details associated with this account will now be: \n \n Username: ${name} \n Email: ${email}` is a message that is being send to user's crypto wallet for signing so it shouldn't be translatable too.

As for missing translations in other components and containers - yes, we want all of them translated but we have new design which will come within a couple of weeks so maybe it's best to skip it - most likely great chunk of texts will be changed there.

bitcoinuser:
> new design which will come within a couple of weeks so maybe it's best to skip it

Where do I see this new design?

soulBit:
Thanks @creed-victor!

creed-victor:
> Where do I see this new design?

@bitcoinuser not sure if it's already there, but all designs should be here: https://drive.google.com/drive/folders/1e_VljWpANJe0o4VmIkKU5Ewo56l9iMaM?usp=sharing

bitcoinuser:
Hi @soulBit, please add this Update button string too. Thanks

Captura de tela de 2021-01-31 15-52-25

Issue #362, opened 2021-02-08
Add visible version number (git commit id) to app


soulBit:
Adding a visible version number somewhere on the app will help debugging issues where the user has not updated to the latest version - e.g:

"Hi, I have a problem.  I want to send BTC to sovryn app, when I go to Fast-BTC-Relay I see the following message:
Address is empty. Anyone who can help me? Thanks"

creed-victor:
Yes, was thinking about too.
I think adding git commit id to footer and console log should do the trick.

soulBit:
@creed-victor good idea! I'll take a stab at this when I have some time this week and get you to review

creed-victor:
> @creed-victor good idea! I'll take a stab at this when I have some time this week and get you to review

Just a note that git commit id should be retrieved in webpack config file and assigned to env variable, maybe something like `REACT_APP_GIT_COMMIT_ID` and printing it in dev console (maybe even from public/index.html file in case react app crashes) and also somewhere in the footer. 
Footer version may even be a hyperlink and open commit in the github when clicked.

soulBit:
@creed-victor Thanks for the pointers, I originally had it copying the commit id to clipboard on click, but your idea is better :)

I'm not sure if the method I'm using to get the commit id in webpack.config,js is optimal, can you check that for me - if there's a way to do it without the package I installed, let's go for that! Same goes for the console.log in public/index.html, I couldn't access the env var with %GIT_COMMIT_ID% like I expected, I assume because it's not loaded from a .env and created by webpack?

If the commit id on the footer is too obtrusive we could also just show the first 7 characters like Github does.

PR #380, opened 2021-02-12
Add git commit id to app for #362

soulBit:


src/app/components/Footer/index.tsx
```diff
@@ -63,6 +64,15 @@ export function Footer() {
...
         </div>
+        <div className="text-center small text-lightGrey">
+          <a
+            href={`https://github.com/DistributedCollective/Sovryn-frontend/commit/${commitHash}`}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            {commitHash}
+          </a>
+        </div>
```
creed-victor, line 75: Maybe add condition so it wouldn't show if there is no commit id? `{commitHash && <div>...</div>}`

src/app/components/Footer/index.tsx
```diff
@@ -63,6 +64,15 @@ export function Footer() {
             </p>
           </div>
         </div>
+        <div className="text-center small text-lightGrey">
+          <a
+            href={`https://github.com/DistributedCollective/Sovryn-frontend/commit/${commitHash}`}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            {commitHash}
```
creed-victor, line 73: Your suggestion in issue is valid, lets keep it short like github does:
```suggestion
            {commitHash.substr(0, 7)}
```

creed-victor:
@kayalight any suggestion how / where to put this so it would look more nicely?

User should be able to find it easily, code in screenshot is too long - will be exactly 7 characters.
It's clickable and I think having it prefixed with something like "Build id"/"Build version" or etc would also help user to find it when asked

Screenshot 2021-02-13 at 21 46 11

kayalight:
what is this exactly communicating, the version of the dapp?
@creed-victor

creed-victor:
> what is this exactly communicating, the version of the dapp?
> @creed-victor

Yes, though "build id" would be more correct term in this case

src/app/components/Footer/index.tsx
soulBit: Good idea, no need to render the container elements if the commit id isn't present

src/app/components/Footer/index.tsx
soulBit: Updated now

soulBit:
@kayalight @creed-victor how about just a simple "build" before the id?

image

Will add the text to translations if you're happy with this

creed-victor:
> @kayalight @creed-victor how about just a simple "build" before the id?
> 
> image
> 
> Will add the text to translations if you're happy with this

Yes, "build" should be ok, just not sure about font and styling, maybe aligning it to the left, adding same styles as other footer texts and making larger space between these elements should do the trick

kayalight:
Screenshot 2021-02-14 at 13 46 17
[build.xd.zip](https://github.com/DistributedCollective/Sovryn-frontend/files/5977802/build.xd.zip)
@creed-victor 
attached is a file with the text styling for the footer section with build ID

kayalight:
[build.xd.zip](https://github.com/DistributedCollective/Sovryn-frontend/files/5977807/build.xd.zip)
@creed-victor

soulBit:
Thanks @kayalight, I like it! If @creed-victor is fine with this, I'll implement it tomorrow

creed-victor:
> Thanks @kayalight, I like it! If @creed-victor is fine with this, I'll implement it tomorrow

Yep, can be implemented

creed-victor (approved):

Issue #389, opened 2021-02-17
Change connected wallet blue circle with address blockie


creed-victor:
![](https://media.discordapp.net/attachments/781522801462083584/811562394290880522/Screenshot_2021-02-17_at_13.38.16.png)

https://github.com/ethereum/blockies

PR #409, opened 2021-02-19
Issue #389 - Change connected wallet blue circle with address blockie

daisai3:


src/app/containers/WalletConnector/index.tsx
```diff
@@ -43,6 +44,18 @@ const WalletConnectorContainer: React.FC<Props> = props => {
...
+    .create({
+      // All options are optional
+      seed: address, // seed used to generate icon data, default: random
+      color: '#dfe', // to manually specify the icon color, default: random
+      bgcolor: '#aaa', // choose a different background color, default: random
+      size: 8, // width/height of the icon in blocks, default: 8
+      scale: 3, // width/height of each block in pixels, default: 4
+      spotcolor: '#000', // each pixel has a 13% chance of being of a third color,
+    })
+    .toDataURL();
```
creed-victor, line 57: I believe it may cause some unnecessary re-renders, putting it to `useEffect` dependency and checking if address is actually set should prevent it

creed-victor (changes requested):

src/app/containers/WalletConnector/index.tsx
daisai3: Since it does not update state, it won't cause some unnecessary re-renders, I think. But if you want I will put it to `useEffect` dependency.

creed-victor: I'm not sure what happens when it uses `address` variable here, if it doesn't cause re-render then all good.
But maybe checking with react dev-tools should make it clear for sure

daisai3: Okay. I'm sure that it doesn't cause re-render, but I will check with react dev-tools.

daisai3: I made a minor change. Please check again.

grinry (approved):

creed-victor (approved):

Issue #415, opened 2021-02-22
Toggable maintenance mode for frontend


creed-victor:

soulBit:
@creed-victor I can work on this one, do you have an idea in mind for styling? Maybe just the Sovryn logo and text below it with "Platform is under maintenance, please try again later" - both in center of the page.

We could also get the design team to create some art of a sleeping badger that we can add to the background, but I'm not sure whether that's "professional" enough for the app.

creed-victor:
This one will also require some backend work, you worked with node.js right?

>We could also get the design team to create some art of a sleeping badger that we can add to the background, but I'm not sure whether that's "professional" enough for the app.

Pretty good idea, maybe @t-z-u team can make something?

----

For task itself, there should be multiple type of maintenance:

- [ ] Full maintenance - change entire dapp with that maintenance "badger"
- [ ] Partial maintenance - only show note in each screen with a text

Maintenance mode should be toggle-able from our admin panel, admin signs message, sends it to backend, backend verifies and saves to database.

When user visits dapp, dapp sends request to backend, retrieves maintenance status and if maintenance set then we showing that full maintenance page or only note/ribbon.

I think that when admin set's maintenance he should be able to add optional maintenance message and choose if maintenance is full (disables entire website) or partial.

Currently we have partial maintenance set by .env and which requires redeployment, it prevents users from making new trades and swaps only so I think it would be perfect if when setting partial maintenance admin would be able to choose which services he wants to disable by checking checkboxes:
- Opening Trades & Swaps
- Opening Loans & Borrows
- Closing Trades
- Closing Loans & Borrows
- Adding & Removing AMM liquidity
- ETH bridge and etc.

If all checkboxes are selected it should be still considered as partial maintenance, so users would be able to see all blockchain data there but they will not be able to make new txes.

soulBit:
> you worked with node.js right?

Yep 👍 

Your solution sounds great, but I don't see an admin panel/db in any of the repos, can you give me permissions?

t-z-u:
@soulBit @creed-victor 
can we concentrate on getting the dapp and design system sorted first?
i count lend, borrow, liquidity, new portfolio section, landing page before we have a full product

soulBit:
@creed-victor this is implemented on `feat/maintenance` branches in db & admin repos, and on `feat/maintenance-toggle` branch in this repo.

Three things to note:

[...]

PR #433, opened 2021-03-02
Feat/maintenance toggle for #415

soulBit:
Supporting changes implemented on `feat/maintenance` branches in db & admin repos


src/app/hooks/useMaintenance.ts
```diff
@@ -0,0 +1,30 @@
...
+  useInjectSaga({ key: maintenanceSlice, saga: maintenanceStateSaga });
+  const dispatch = useDispatch();
+  const store = useSelector(selectMaintenanceState);
+
+  useEffect(() => {
+    if (!store || Object.keys(store).length === 0) {
+      dispatch(actions.fetchMaintenanceState());
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
creed-victor, line 23: Thinking that maybe we should refetch state from the backend every minute or so, otherwise users who already in the dapp will be able to make transactions and pay fee for failed transactions.

You can do it in saga:

```typescript
import delay from '@redux-saga/delay-p'

function* fetchStatePeriodicallySaga() {
  while (true) {
    yield call(callCreateMaintenanceState);
    yield delay(2000);
  }
}
```

creed-victor (changes requested):

creed-victor (approved):

PR #478, opened 2021-03-16
Feat/460 light charts

### LINKED ISSUE #460: Implement lightweight-charts for our trading pairs
Currently we are using tradingview widget as price history chart but it retrieves data from exchange, so price is not same as in our systems and they don't have same pairs we have.

Idea is to swap widget with https://github.com/tradingview/lightweight-charts library to get price history from our backend.

- when chart component is mounted it should request backend for history from current time to 3 month old
- every x seconds (make it prop and 15 by default) request backend for new price data. Limit it to last retrieved date until current date so we wouldn't need to retrieve all 3 month data from database each time.
- it should be candle chart, if it can be changed to another type by prop even better.
- pair examples BTC/DOC, BTC/BPRO, BTC/USDT, BPRO/USDT

This task also requires backend work on Sovryn-database repository:

- collect prices for pairs from blockchain to database table
- endpoint for computing and showing this data as json, filtered by pair and from / to dates (timestamps)
- make sure pairs in database repository are easily defined in one place and can be reused for both collecting prices from blockchain to write to database and retrieving it from database and printing in endpoint.

I think it's best to reach out @BetsyBraddock when working with it as backend so she would show you the way on how to get prices, it's even possible that prices are already somewhere in the database.

soulBit:
Adds lightweight-trading-charts support to Trading page. Closes #460


src/app/components/TradingChart/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+                  break;
+                }
+              if (i > -1) newSeries = newSeries.slice(0, i);
+            }
+            newSeries = newSeries.concat(response.data.series);
+            // console.log(newSeries);
+            setChartData([{ data: newSeries }]);
+            const latest = newSeries[newSeries.length - 1];
+            setLastTime(latest.time * 1e3); // datum time is in seconds, lastTime is ms
+            console.log(latest.time, latest.close);
```
creed-victor, line 90:
```suggestion
```

src/app/components/TradingChart/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+                  i = x;
+                  break;
+                }
+              if (i > -1) newSeries = newSeries.slice(0, i);
+            }
+            newSeries = newSeries.concat(response.data.series);
+            // console.log(newSeries);
+            setChartData([{ data: newSeries }]);
+            const latest = newSeries[newSeries.length - 1];
+            setLastTime(latest.time * 1e3); // datum time is in seconds, lastTime is ms
```
creed-victor, line 89: When backend returns empty `series` array chart fails to load and console shows this error `TypeError: Cannot read property 'time' of undefined`

Checking if response.data.series.length is not 0 here too should do the trick

creed-victor (changes requested):

creed-victor:
Also, maybe change chart background to #171717 and if possible add white border.
Tzu will have to prepare some colors for it in the future but for now new background should be ok

creed-victor:
Another request - can we have legend like in the screenshot (the one with date and price which changes when cursor is moving) below pair name?

Screenshot 2021-03-17 at 09 19 20

Code of implementation:
https://jsfiddle.net/TradingView/dtehqksw/

soulBit:
I've added all the requested features now - for the onHover price display in candle mode, I've shown each of the available props (`open`. `low`, `high`, `close`) instead of just the `close` price like the example you provided - screenshot of both to compare:

image

If you prefer just the close value, it's easily changed in `utils/kaktana-lightweight-charts/Chart.js::311`

creed-victor (approved):

Issue #450, opened 2021-03-06
[YBUG.IO] Hi, I found this UX minor issue...


creed-victor:
Hi, I found this UX minor issue that when a user tries to swap (eg: BTC -> USDT) and then if the user wants to make the inverse swap (USDT -> BTC), the user is not able to select the coin without having to select a 3rd coin to enable the desired option the coin select field. An easy solution could be to use an arrow between both field to easy invert the selection.

*Source url:* https://live.sovryn.app/
*Reported by:* mrafaelgomes@gmail.com
*Reported at:* 06 Mar at 22:52 UTC

*Location:* PT, Aveiro, Aveiro
*Browser:* Chrome 88.0.4324.192

*OS:* macOS 11.2.2
*Screen:* 1440x900
*Viewport:* 1440x764

*Screenshot:*
![Screenshot](https://ybug.io/data/reports/3f1jrxvzrhkn1b975t8b/t3gdt5041jg3knwtk9yp/screenshot.jpg?_token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2MTUwNzExNjgsImRhdGEiOnsicmVwb3J0IjoidDNnZHQ1MDQxamcza253dGs5eXAiLCJ1c2VyIjoiR2l0SHViIn19._IlIRYNKwUiUH7m7hK99m2tJQzO2nEBIPk_eTSWS8Cs)

For more details please visit report page on Ybug:
[https://ybug.io/dashboard/reports/detail/t3gdt5041jg3knwtk9yp](https://ybug.io/dashboard/reports/detail/t3gdt5041jg3knwtk9yp)

creed-victor:
@soulBit maybe you will find some time to take on this one?

soulBit:
@creed-victor thanks, will take a look

creed-victor:
I think easiest fix would be to just disable submit button if target = source asset

creed-victor:
Just not sure how amount for "you receive" will be returned, most likely smart-contract will return error so maybe show `0` or `-` when that happens

PR #484, opened 2021-03-19
Implement Swap Form switch assets button for #450

soulBit:
@creed-victor it will reset amount to 0 when switching tokens, since user most likely will need to adjust balance anyway - see what you think.

I also set BTC -> DOC as default pair


grinry (approved):

creed-victor (approved):
Cool, thanks

PR #493, opened 2021-03-21
Update Lend/borrow header link text to "Loans". Closes #485

### LINKED ISSUE #485: Rename Lend/Borrow to Loans
Hi,

I think the button Lend/Borrow should be renamed to Loans.

Why?
Because when translating Lend/Borrow to Portuguese, the string stay very long and it's "ugly". Will be Emprestar/Tomar Empréstimo.

Captura de tela de 2021-03-20 12-37-48

soulBit:
Closes #485

### DIFF
--- a/src/locales/pt_br/translation.json
+++ b/src/locales/pt_br/translation.json
@@ -150,7 +150,7 @@
   },
   "mainMenu": {
     "trade": "Operar",
-    "lend": "Empréstimos",
+    "lend": "Loans",
     "fastBtc": "Depositar-Btc",
     "wallet": "Portfólio",
     "liquidity": "Liquidez",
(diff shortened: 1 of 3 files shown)


src/locales/pt_br/translation.json
```diff
@@ -150,7 +150,7 @@
   },
   "mainMenu": {
     "trade": "Operar",
-    "lend": "Empréstimos",
+    "lend": "Loans",
```
bitcoinuser, line 153:
```suggestion
    "lend": "Empréstimos",
```

bitcoinuser (changes requested):
In PT-BR was correct. No need to change.

creed-victor (approved):

PR #502, opened 2021-03-23
Feat/show arbitrage opportunity

BetsyBraddock:
This is the same PR as the one to development for this feature:
image

I just read the ReadMe and realised I should have made the PR to master - thanks!


creed-victor:
Only fixes goes to master branch, I consider this one as a new feature so PR to development branch was correct one :)

If feature must be released as soon as possible then PR can go directly to the master, but branch must be branched out of the master too - otherwise there is possibility to release features we didn't intend to

PR #518, opened 2021-03-24
bug/512-remove0

boltsr:
change the decimal of max amount


creed-victor:
I think idea here is to show only 2 numbers after decimal point, but only if trailing numbers are zeros.
so `0.2500` becomes `0.25`, `0.1000` -> `0.10`, `0.1230` -> `0.123`, `0.1234` -> `0.1234`, `0.12345` -> `0.1234`, `0.0010` -> `0.001` -> `0.0015` -> `0.0015`. 

It should be applied for both max and min amount, these numbers are retrieved from the backend so it can change thus meaning there should be a possibility to have show 4 decimals.

Another part of this task is to keep BTC symbols for both min and max values aligned in one line. Having amounts in one div,  BTC symbols in another and making them as a row items should do the trick.

boltsr:
I see your point

creed-victor (approved):

PR #533, opened 2021-03-25
Feat/501 asset refactor

### LINKED ISSUE #501: Change enum Asset.BTC to Asset.RBTC
Enum asset has BTC property which is actually incorrect - property, value and it's usages should be changed to RBTC.

File:
`./src/types/assets.ts`

soulBit:
DistributedCollective/Sovryn-database/pull/78 needs to be merged & deployed before this PR

Closes #501

### DIFF
--- a/src/app/containers/SwapTradeForm/components/arbitrage.tsx
+++ b/src/app/containers/SwapTradeForm/components/arbitrage.tsx
@@ -1,6 +1,5 @@
 import React, { useState, useEffect } from 'react';
 import { backendUrl, currentChainId } from '../../../../utils/classifiers';
-import { Asset } from '../../../../types/asset';
 import axios from 'axios';
 import { symbolByTokenAddress } from 'utils/blockchain/contract-helpers';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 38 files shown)


src/app/containers/SwapTradeForm/components/arbitrage.tsx
```diff
@@ -31,9 +32,9 @@ export function Arbitrage() {
     // Only show component if you can earn > 0.001 BTC or > 1 USDT
     if (
       (data.USDT.rateToBalance.earn > 0.001 &&
-        symbolByTokenAddress(data.USDT.rateToBalance.to) === 'BTC') ||
+        symbolByTokenAddress(data.USDT.rateToBalance.to) === Asset.RBTC) ||
```
creed-victor, line 35: @BetsyBraddock this data goes from backend and must be `BTC`, `RBTC` would fail assertion here?

BetsyBraddock: Yes you are right - backend uses BTC token abbreviation instead of RBTC

soulBit: Thanks, this is updated now

PR #548, opened 2021-03-29
Changed stats page to include tvl and trading volume

BetsyBraddock:
This is two new components for the "Stats" page of the dapp, showing Trading Volume and TVL,. It looks like this: 
image


creed-victor:
@t-z-u something to consider in stats page redesign

PR #579, opened 2021-04-04
bug/558-fastbtc-buttons

boltsr:
fix fastBTC dialog button width


bitcoinuser:
Hi @boltsr you fixed the problem on fastbtc window but now on wallet page the buttons are with large width. And the action is not aligned too.

Captura de tela de 2021-04-04 14-20-23

creed-victor (approved):

PR #582, opened 2021-04-04
Feat/liquidity mining

BetsyBraddock:
A few small improvements to the liquidity mining UI to show both rewards pools even if the user has only deposited into one asset pool


src/app/containers/LiquidityMining/index.tsx
```diff
@@ -73,21 +74,52 @@ export function LiquidityMining() {
...
         Liquidity Mining: USDT/BTC Pool
       </h1>
       <div className="d-flex flex-wrap mb-5"></div>
-      <div className="d-flex flex-wrap justify-content-around">{rows}</div>
+      <div className="d-flex flex-wrap justify-content-around">
+        <PoolData
+          data={
+            USDTData || {
+              asset: getContract('USDT_token').address,
+              pool: getContract('USDT_amm').address,
```
creed-victor, line 100: Would recommend to use getTokenContract and getAmmContract helpers instead, in case name properties would change here.

creed-victor (approved):

PR #627, opened 2021-04-12
Add and translate strings to PT-BR

bitcoinuser:


bitcoinuser:
Hi @creed-victor please merge this. I know has some strings not translated yet, but will take some time to complete. But I think is important to have this merged even with some strings not translated yet.

PR #739, opened 2021-04-15
Feat/swap slippage

rick23p:


src/app/containers/SwapFormContainer/index.tsx
```diff
@@ -210,15 +218,13 @@ export function SwapFormContainer() {
...
-            <span>Maximum Received: 0.02931216</span>
-            <span>
-              <img
-                src={settingIcon}
-                alt="settings"
-                onClick={() => setDialogOpen(true)}
-              />
-            </span>
+          <div className="swap-btn-helper tw-flex tw-items-center tw-justify-center">
+            <span>Minimum Received: {weiToNumberFormat(minReturn, 8)}</span>
```
creed-victor, line 222: Add translation key for this string

src/app/containers/SwapFormContainer/index.tsx
```diff
@@ -210,15 +218,13 @@ export function SwapFormContainer() {
...
+            <img
+              src={settingIcon}
+              alt="settings"
+              onClick={() => setDialogOpen(true)}
+            />
           </div>
           <button
```
creed-victor, line 229: Change button with component that was used in BuySovPage form, to make sure it has all states (hover, disabled, etc)

If user is not engaged wallet, button should say "Engage Wallet" - opens engage wallet dialog when clicked

creed-victor (changes requested):

creed-victor:
Remove this component, we are showing progress in the dialog now
Screenshot 2021-04-16 at 09 32 35

Change account balance with component from BuySOVPage (where everything is in one line)
Change amount field to /app/components/form/AmountInput component

Screenshot 2021-04-16 at 09 36 30

creed-victor (changes requested):
Clone TxDialog component from BuySovPage to global components and make sure that clicking "confirm" button in cloned dialog doesn't redirect user to portfolio page, it should just close that dialog.

creed-victor (approved):

Issue #755, opened 2021-04-16
Header changes


creed-victor:
- [ ] create new dropdown group named `Bitocracy`
- [ ] move `Stake` and `Governance` navlinks to `Bitocracy` dropdown
- [ ] Make Stake and Governance dropdowns open in the new tab again
- [ ] Add icon which means that link will open in the new tab (ask tzu to make one)
- [ ] Do same thing for bitocracy.sovryn.app
- [ ] Open dropdowns on hover instead of click

bitcoinuser:
Two more suggestions:

1- Just need to hover the mouse (no need to click) to show the other buttons from dropdown.
2- Add descriptions like the binance.com dropdown menu.

bitcoinuser:
This kind of menu from gate.io site is interesting:

![Captura de tela de 2021-04-18 23-43-33](https://user-images.githubusercontent.com/1178856/115174710-20c62480-a0a0-11eb-9a51-4469c20a405a.png)

bitcoinuser:
The menu still is not good on PT-BR on https://test.sovryn.app/.

![Captura de tela de 2021-04-21 14-35-04](https://user-images.githubusercontent.com/1178856/115597423-a23fd180-a2af-11eb-9f36-2ffe6340d9c8.png)

PR #823, opened 2021-04-20
[Header] - added fixes for #755

pietro-maximoff:
- create new dropdown group named Bitocracy
- move Stake and Governance navlinks to Bitocracy dropdown
- Make Stake and Governance dropdowns open in the new tab again
- Add icon which means that link will open in the new tab
- Do same thing for bitocracy.sovryn.app
- Open dropdowns on hover instead of click

### DIFF
--- a/src/locales/pt_br/translation.json
+++ b/src/locales/pt_br/translation.json
@@ -202,7 +202,8 @@
     "governance": "Governança",
     "wallet": "Portfólio",
     "stats": "Dados",
-    "help": "Ajuda"
+    "help": "Ajuda",
+    "bitocracy": "Bitocracy"
   },
   "trandingPositionSelector": {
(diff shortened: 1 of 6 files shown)


src/locales/pt_br/translation.json
```diff
@@ -202,7 +202,8 @@
     "governance": "Governança",
     "wallet": "Portfólio",
     "stats": "Dados",
-    "help": "Ajuda"
+    "help": "Ajuda",
+    "bitocracy": "Bitocracy"
```
bitcoinuser, line 206:
```suggestion
    "bitocracy": "Bitocracia"
```

bitcoinuser (changes requested):
Change Pt-br

creed-victor (approved):

PR #913, opened 2021-04-26
Improve opt-out link

Bitcoinera:
Make the link for the opt-out also work for non-Chromium-base browsers


Bitcoinera:
OK, tested and working :+1:

creed-victor (approved):

PR #955, opened 2021-04-28
Escrow

creed-victor:


creed-victor:
TODO:

- [ ] check if smart-contract status = Deposit, otherwise disable deposit button

PR #963, opened 2021-04-29
Staking page

pietro-maximoff:


src/app/components/Header/index.tsx
```diff
@@ -147,10 +147,7 @@ export function Header() {
       },
     },
     { to: '/liquidity', title: t(translations.mainMenu.liquidity) },
-    {
-      to: 'https://bitocracy.sovryn.app/stake',
-      title: t(translations.mainMenu.staking),
-    },
+    { to: '/stake', title: t(translations.mainMenu.staking) },
```
creed-victor, line 150: I think it still should be under a dropdown group, but not sure if under bitocracy one, maybe under finance or somewhere in the portfolio itself?
@t-z-u

src/app/components/LinkToExplorer/index.tsx
```diff
@@ -50,7 +51,7 @@ export function LinkToExplorer(props: Props) {
   return (
     <a
       className={props.className}
-      href={`${url}/tx/${props.txHash}`}
+      href={`${url}/${props.isAddress ? 'address' : 'tx'}/${props.txHash}`}
```
creed-victor, line 54: I think we should have another component for addresses, like could be named `AddressBadge`

src/store/global/transactions-store/types.ts
```diff
@@ -12,6 +12,10 @@ export interface TransactionsStoreState {
 export type ContainerState = TransactionsStoreState;
 
 export enum TxType {
+  STAKE = 'stake',
+  WITHDRAW = 'withdraw',
+  EXTEND = 'extend',
+  DELEGATE = 'delegate',
```
creed-victor, line 18: Lets add STAKING_ prefix to prevent duplicates and make sure other devs will use correct type

```suggestion
  STAKING_STAKE = 'staking-stake',
  STAKING_WITHDRAW = 'staking-withdraw',
  STAKING_EXTEND = 'staking-extend',
  STAKING_DELEGATE = 'staking-delegate',
```

src/utils/blockchain/contracts.testnet.ts
```diff
@@ -30,6 +31,15 @@ export const contracts = {
...
   },
+  sovToken: {
+    address: '0x6a9A07972D07e58F0daf5122d11E069288A375fb',
+    abi: SovTokenABI,
+    blockNumber: 2742422,
+  },
+  vestingRegistry2: {
+    address: '0x310006E356b0818C3Eaf86a9B2f13013d4691a1c',
+    abi: VestingRegistryAbi as any,
+  },
```
creed-victor, line 42: SOV token contract is already defined and called `SOV_token`, you will need to use this name from now one.
Vesting registry 2 is called `vestingRegistryOrigin`

```suggestion
```

src/utils/blockchain/contracts.ts
```diff
@@ -30,6 +31,15 @@ export const contracts = {
...
   },
+  sovToken: {
+    address: '0xEFc78fc7d48b64958315949279Ba181c2114ABBd',
+    abi: SovTokenABI,
+    blockNumber: 2742422,
+  },
+  vestingRegistry2: {
+    address: '0x0a9bDbf5e104a30fb4c99f6812FB85B60Fd8D372',
+    abi: VestingRegistryAbi as any,
+  },
```
creed-victor, line 42:
```suggestion
```

src/utils/sovryn/contract-writer.ts
```diff
@@ -138,21 +139,26 @@ class ContractWriter {
...
   ): Promise<string | RevertInstructionError> {
+    let params = args;
+    if (args && args.length && typeof args[args.length - 1] === 'object') {
+      params = args.slice(0, -1);
+    }
     return new Promise<string | RevertInstructionError>(
       async (resolve, reject) => {
         const data = this.getCustomContract(address, abi)
-          .methods[methodName](...args)
+          .methods[methodName](...params)
```
creed-victor, line 161: Not sure if this will not break other functionality, will need to make some tests.

creed-victor (changes requested):
Date selector style is broken:

Screenshot 2021-04-29 at 19 23 22

creed-victor (changes requested):
Sorry, added comment to wrong PR.

pietro-maximoff:
> Date selector style is broken:
> 
> Screenshot 2021-04-29 at 19 23 22

maybe you have some errors in console regarding slider? because i can't catch it, tested in chrome and ff

creed-victor:
> maybe you have some errors in console regarding slider? because i can't catch it, tested in chrome and ff
slider is a library perhaps? it's possible that tailwind purger is removed styles for it's classes because it thinks that classes are not used.
In this case this issue is visible only when built

src/utils/blockchain/contracts.testnet.ts
pietro-maximoff: done

src/utils/blockchain/contracts.ts
pietro-maximoff: done

src/store/global/transactions-store/types.ts
pietro-maximoff: done

pietro-maximoff:
> > maybe you have some errors in console regarding slider? because i can't catch it, tested in chrome and ff
> 
> @pietro-maximoff slider is a library perhaps? it's possible that tailwind purger is removed styles for it's classes because it thinks that classes are not used.
> In this case this issue is visible only when built

yes, right. i checked on prod build and have the same issue, i'll rewrite all styles for slider and it will be fixed

pietro-maximoff:
> > > maybe you have some errors in console regarding slider? because i can't catch it, tested in chrome and ff
> > 
> > 
> > @pietro-maximoff slider is a library perhaps? it's possible that tailwind purger is removed styles for it's classes because it thinks that classes are not used.
> > In this case this issue is visible only when built
> 
> yes, right. i checked on prod build and have the same issue, i'll rewrite all styles for slider and it will be fixed

done

src/app/hooks/staking/useSoV_balanceOf.ts
```diff
@@ -0,0 +1,11 @@
+import { useCacheCallWithValue } from '../useCacheCallWithValue';
+import { ethGenesisAddress } from '../../../utils/classifiers';
+
```
creed-victor, line 3: Hook can be deprecated and `useAssetBalanceOf` instead.

src/utils/helpers.ts
```diff
@@ -172,3 +173,58 @@ export function makeId(length: number = 8): string {
...
+    default:
+      throw new Error('Unsupported unit (custom toWei helper)');
+  }
+
+  return roundToSmaller(bignumber(amount || '0').mul(10 ** decimals), 0);
+};
+
+export const numberFromWei = (amount: any, unit: Unit = 'ether') => {
+  return Number(fromWei(amount, unit));
+};
```
creed-victor, line 224: I think these functions should be already in the codebase with same name, expect for numberFromWei - it may have another name

src/utils/sovryn/contract-writer.ts
creed-victor: Looks like other dapp parts didn't broke, so should be ok

creed-victor (changes requested):
Looks like everything is working now, but lets not merge it to the dev until redesign features are on mainnet.

While waiting, couple of improvements:
1. When sending transaction we should show `TxDialog` component which instructs users to check on their wallet for instructions (otherwise user doesn't see any response from the dapp when using hardware wallet).
2. Change staking history to get events from the database instead of blockchain (dapp has a hook for that)
3. There is couple of functions and hooks mentioned in previous comments that is duplicates - remove them and use the ones that was already in the dapp
4. Numbers should be wrapped with `toNumberFormat` or `weiToNumberFormat` functions to transform numbers to localised format (for example: instead of 1000000 users should see 1,000,000.00)
5. Add translation keys to the strings

PR #1008, opened 2021-05-03
Feat/959 pending transaction in tables.

rick23p:


src/app/containers/SwapHistory/index.tsx
```diff
@@ -219,16 +257,24 @@ function AssetRow({ data, itemFrom, itemTo }: AssetProps) {
       <td>
         <div className="d-flex align-items-center justify-content-between col-lg-10 col-md-12 p-0">
           <div>
-            {tx && <p className="m-0">Confirmed</p>}
-            {!tx && <p className="m-0">Failed</p>}
+            {!data.status && <p className="m-0">Confirmed</p>}
+            {data.status === TxStatus.FAILED && <p className="m-0">Failed</p>}
+            {data.status === TxStatus.PENDING && <p className="m-0">Pending</p>}
```
creed-victor, line 262: Add translations to this, I think state translation keys are defined in common group

src/app/pages/MarginTradePage/components/OpenPositionsTable/index.tsx
```diff
@@ -73,6 +77,19 @@ export function OpenPositionsTable(props: Props) {
...
             </tr>
           )}
+          {transactions.length > 0 && (
+            <>
+              {transactions
+                .filter(
+                  tx =>
+                    tx.type === TxType.TRADE &&
+                    [TxStatus.FAILED, TxStatus.PENDING].includes(tx.status),
+                )
```
creed-victor, line 87: I think filter here will cause rerender if unrelated transaction will be added / updated to the transactions. Maybe filtering should be done in the memo?

creed-victor (changes requested):
1. make sure latest pending transaction is first in the list (order desc). 
2. when opening short position it shows position size as rbtc, but should actually be as rusdt - see screenshot


Screenshot 2021-05-03 at 14 06 42

creed-victor (approved):

PR #1009, opened 2021-05-03
Spot trading UI

creed-victor:


bitcoinuser:
@creed-victor What is the difference between Spot trade and swap?

creed-victor:
> @creed-victor What is the difference between Spot trade and swap?

currently it's same thing just in different clothes, has charts - swap page doesn't anymore.

creed-victor:
In history add <AssetRenderer /> component for showing asset names (it transforms RBTC to rBTC (r with opacity)) :)

PR #1090, opened 2021-05-12
fix stringToFixedPrecision issue.

### LINKED ISSUE #1089: allow manual entry of price
on the margin trading page user cannot type in the amount

rick23p:
closes #1089

### DIFF
--- a/src/utils/display-text/format.ts
+++ b/src/utils/display-text/format.ts
@@ -112,8 +112,7 @@ export function calculateProfit(
 export const countDecimals = value => {
   if (!value) return;
   const decimalPart = value.split('.')[1];
-  if (Math.floor(value) === value || !decimalPart) return 0;
-  return decimalPart.length || 0;
+  return decimalPart ? decimalPart.length || 0 : 0;
 };
 
 export const stringToFixedPrecision = (value: string, precision: number) => {
--- a/src/utils/display-text/format.ts
+++ b/src/utils/display-text/format.ts
@@ -109,14 +109,13 @@ export function calculateProfit(
   return isLong ? profitLong : profitShort;
 }
 
-export const countDecimals = value => {
-  if (!value) return;
-  const decimalPart = value.split('.')[1];
-  return decimalPart ? decimalPart.length || 0 : 0;
+const decimalPartLength = value => {
+  const decimalPart = value?.split('.')[1];
+  return decimalPart?.length ?? 0;
 };
 
 export const stringToFixedPrecision = (value: string, precision: number) => {
-  if (countDecimals(value) > precision)
+  if (decimalPartLength(value) > precision)
     return parseFloat(value).toFixed(precision);
 
   return value;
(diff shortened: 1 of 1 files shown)


src/utils/display-text/format.ts
```diff
@@ -109,5 +109,13 @@ export function calculateProfit(
   return isLong ? profitLong : profitShort;
 }
 
-export const stringToFixedPrecision = (value: string, precision: number) =>
-  parseFloat(value).toFixed(precision);
+export const countDecimals = value => {
```
tiltom, line 112: You don't need to export this function since the only usage is in the same file (and I don't think we would need it elsewhere). 

Also, rename it to `decimalPartLength` please.

src/utils/display-text/format.ts
```diff
@@ -109,5 +109,13 @@ export function calculateProfit(
...
+  if (Math.floor(value) === value || !value.split('.')[1]) return 0;
```
tiltom, line 113: I would extract `value.split('.')[1]` to a variable (e.g. `decimalPart`) and use it here as `!decimalPart`.

src/utils/display-text/format.ts
```diff
@@ -109,5 +109,13 @@ export function calculateProfit(
...
+  return value.split('.')[1].length || 0;
```
tiltom, line 114: The OR part is unnecessary here since you do all the checks before + you can use newly created variable as `decimalPart.length`.

src/utils/display-text/format.ts
```diff
@@ -109,5 +109,13 @@ export function calculateProfit(
...
+};
+
+export const stringToFixedPrecision = (value: string, precision: number) => {
+  if (countDecimals(value) > precision)
+    return parseFloat(value).toFixed(precision);
+  else return value;
```
tiltom, line 120: Unnecessary `else`, you can just return.

tiltom (changes requested):

src/utils/display-text/format.ts
tiltom: Why you didn't address this comment? Let me know if you have another opinion on it but I think those are valid arguments.

src/utils/display-text/format.ts
```diff
@@ -109,5 +109,15 @@ export function calculateProfit(
   return isLong ? profitLong : profitShort;
 }
 
-export const stringToFixedPrecision = (value: string, precision: number) =>
-  parseFloat(value).toFixed(precision);
+export const countDecimals = value => {
+  if (!value) return;
```
tiltom, line 113: You don't need this, just use `value?.split('.')[1];` on the next line.

src/utils/display-text/format.ts
```diff
@@ -109,5 +109,15 @@ export function calculateProfit(
...
+  const decimalPart = value.split('.')[1];
+  return decimalPart ? decimalPart.length || 0 : 0;
```
tiltom, line 115: You could rewrite it to `return decimalPart?.length ?? 0`.

tiltom (changes requested):

tiltom (approved):

t-z-u (approved):

PR #1092, opened 2021-05-12
Update open positions and trading history table on Margin trade page

tiltom:


bitcoinuser:
Hi @tiltom, this should be aligned, no?

Captura de tela de 2021-05-12 11-48-18

tiltom:
> Hi @tiltom, this should be aligned, no?
> 
> Captura de tela de 2021-05-12 11-48-18

Hi, I see it aligned:
image

However, those are 2 different tables with a different set of columns so there is no guarantee it will be aligned and I cannot force it (technically I could with fixed columns width but it's not pretty and it's unnecessary in this case).

creed-victor (approved):

pietro-maximoff (approved):

bitcoinuser:
> > Hi @tiltom, this should be aligned, no?
> > Captura de tela de 2021-05-12 11-48-18
> 
> Hi, I see it aligned:
> image
> 
> However, those are 2 different tables with a different set of columns so there is no guarantee it will be aligned and I cannot force it (technically I could with fixed columns width but it's not pretty and it's unnecessary in this case).

Hi @tiltom I asked because see this in PT-BR language:

Captura de tela de 2021-05-14 11-14-11

PR #1093, opened 2021-05-12
feat/update-notification-dialog

boltsr:
update dialog


creed-victor (changes requested):
Submitting fails saying that var `eth` can't be found in undefined.
That's because `Sovryn.getWriteWeb3().eth.personal` shouldn't be used anymore (we moved connection to `@sovryn/wallet` and getWeiteWeb3 is not connected in our case)

Instead you should use 
```
import { walletService } from '@sovryn/wallet';

walletService.signMessage(message);
```

creed-victor (approved):

pietro-maximoff (approved):

PR #1096, opened 2021-05-13
Allow manual entry on Swap page

tiltom:


pietro-maximoff (approved):

creed-victor (changes requested):
Lets rework this and implement to amount component itself, so amount input would behave exactly the same in all pages

creed-victor (approved):

pietro-maximoff (approved):

PR #1107, opened 2021-05-14
Finance V2 shared components

tiltom:


src/app/components/FinanceV2Components/TableTransactionStatus/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const TableTransactionStatus: React.FC<ITableTransactionStatusProps> = ({
+  iconClassName,
+  textClassName,
+  transactionStatus,
+  wrapperClassName,
+}) => {
+  const { t } = useTranslation();
+
+  const getText = useCallback(() => {
```
creed-victor, line 25: I think these could use `useMemo` instead of `useCallback`

src/app/components/Footer/index.tsx
```diff
@@ -84,6 +84,18 @@ export function Footer() {
...
+          <a
+            href="https://www.highcharts.com"
+            target="_blank"
+            rel="noopener noreferrer"
+            className="tw-font-normal tw-text-white"
+          >
+            Highcharts
+          </a>
+          . All rights reserved.
+        </div>
```
creed-victor, line 98: Eh, it's requirement of highcharts event when we have license?

src/app/pages/LiquidityMining/components/HistoryTable/TableBody/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { TableRow } from '../TableRow/index';
+
+export const TableBody: React.FC = () => {
+  const pools = LiquidityPoolDictionary.list();
+
+  return (
+    <tbody className="mt-5">
+      {pools.map((item, index) => (
+        <TableRow
+          key={`${item.supplyAssets[0].asset}/${item.supplyAssets[1].asset}/${index}`}
```
creed-victor, line 12: If i understand correctly, you can shorten it by using item.poolAsset (it should be unique for each pool, at least for now)
```suggestion
          key={`${item.poolAsset}/${index}`}
```

src/app/components/FinanceV2Components/RowTable/TableBody/index.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+export const TableBody: React.FC = ({ children }) => (
+  <tbody className="tw-border-t tw-border-b">
+    <tr className="tw-h-16">{children}</tr>
+  </tbody>
+);
```
creed-victor, line 7: `tr` here means that table must have only one row, that's correct?

src/app/components/Footer/index.tsx
tiltom: Well, I don't know, but maybe @soulBit knows more.

src/app/pages/LiquidityMining/components/HistoryTable/TableBody/index.tsx
tiltom: I wanted to be super sure it won't clash but it's true that it looks bit overengineered :D

src/app/components/FinanceV2Components/RowTable/TableBody/index.tsx
tiltom: Exactly but in this case it's intentional.

creed-victor (approved):

PR #1151, opened 2021-05-22
Restyle updates

soulBit:
Restyle updates from Tzu

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -26,6 +26,20 @@ module.exports = {
     fontFamily: {
       body: ['Montserrat', 'sans-serif'],
     },
+    fontSize: {
+      tiny: '.7rem',
+      xs: '.75rem',
+      sm: '.875rem',
+      base: '1rem',
+      lg: '1.125rem',
+      xl: '1.25rem',
+      '2xl': '1.5rem',
+      '3xl': '1.875rem',
+      '4xl': '2.25rem',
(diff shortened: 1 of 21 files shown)


t-z-u (approved):

tailwind.config.js
```diff
@@ -26,6 +26,20 @@ module.exports = {
...
+      sm: '.875rem',
+      base: '1rem',
+      lg: '1.125rem',
+      xl: '1.25rem',
+      '2xl': '1.5rem',
+      '3xl': '1.875rem',
+      '4xl': '2.25rem',
+      '5xl': '3rem',
+      '6xl': '4rem',
+      '7xl': '5rem',
```
grinry, line 41: Step size is 0.25rem (4px), would like to see some naming convention for fonts like its for the widths/heights in the future too so we could use classes without checking which size it defines everytime.
For example tw-w-4 is 16px, tw-w-8 is 32px and so on (considering base font size is 16px)

grinry (approved):

tailwind.config.js
soulBit: Agreed, I just added `tiny` here for now and the rest are the standard TW sizes, but having our own predefined names/sizes works better like you mentioned.

Once we've finished the new redesign pages, Tzu wants to make sure all the dapp components and text we are using are consistent, so we can define sizes at that point and remove existing ones

Issue #870, opened 2021-04-22
Add confirmation modal to the end of Engage wallet flow


t-z-u:

t-z-u:
![Uploading 15 – 58@2x.png…]()

t-z-u:
![15 – 58@2x](https://user-images.githubusercontent.com/79344165/115789983-e3bea280-a3bd-11eb-9792-867df850d10b.png)

tiltom:
Could you provide a prototype link @t-z-u ?

PR #1154, opened 2021-05-24
change AddressQrCode copy success.

### LINKED ISSUE #1149: add address copied to fastbtc modal

rick23p:
closes #1149


soulBit (approved):

grinry (approved):

grinry:
Things like this should have their own component and reused (I think there was multiple places where copy-to-clipboard was used), but ok.

tiltom:
> Things like this should have their own component and reused (I think there was multiple places where copy-to-clipboard was used), but ok.

You're right + we have another story where we need to use it: https://github.com/DistributedCollective/Sovryn-frontend/issues/870 .

creed-victor (approved):

PR #1155, opened 2021-05-24
Automatically update liquidity on successful transactions

tiltom:


tiltom:
The diff looks a bit crazy but I just basically moved down the `useEffect` and added some more dependencies to it so the balances refresh when other values change.

pietro-maximoff (approved):

grinry (approved):

PR #1173, opened 2021-05-25
Origins launchpad - no sale on

tiltom:


tiltom:
The code is ready for review, the only question is if we will have email integration ready by the launch as this wait for backend. That's why there is the email input with handler that does nothing, it will be either commented-out or finished but that does not block the review.

src/app/pages/OriginsLaunchpad/components/PreviousSalesCardSection/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { PreviousSalesCard } from '../PreviousSalesCard/index';
+import imgGenesis from 'assets/origins_launchpad/GEN-NFT-COMMUNITY.svg';
+import imgOrigins from 'assets/origins_launchpad/GURU_BADGER_NFT.svg';
+
+export const PreviousSalesCardSection: React.FC = () => (
+  <PreviousSalesRowWrapper>
+    <PreviousSalesRow>
+      <PreviousSalesCard
+        saleName="SOV Origins Sale"
+        saleAllocation="2,000,000 SOV"
```
creed-victor, line 12: Numbers should use `Number.toLocaleString` prototype or to `toNumberFormat` helper for displaying, hardcoding format will make only couple of regions to understand what is the actual number here.

creed-victor: And currently displaying number formats are also different for each cards, here it's 2,000,000 and in genesis sale card it's 2641946.1868

creed-victor (changes requested):
Looks good except for 2 things:
1. Inconsistent number format displaying (always use toNumberFormat helper or toLocaleString prototype so it would be clear what's the actual number for regions worldwide, not only UK)
2. Header has too much of nav elements already, @t-z-u as someone already suggested - i think it's time to move non-trading related stuff to "others" like binance has for example:
Screenshot 2021-05-31 at 10 53 16

src/app/pages/OriginsLaunchpad/components/PreviousSalesCardSection/index.tsx
tiltom: Good catch, I'll fix the numbers and regarding the menu - it's more the question for @t-z-u.

tiltom: The numbers format issue is fixed.

tiltom:
Since we want to launch it today and the development branch is not yet ready to be merged to master, I squashed all of the commits into one and cherry-picked it to a new branch that has master as a base. 

I am going to close this PR and https://github.com/DistributedCollective/Sovryn-frontend/pull/1228 is the new Origins launchpad PR.

PR #1176, opened 2021-05-25
feat/1079_borrow_v2

boltsr:
Implement borrow v2


creed-victor (changes requested):
`public/datafeeds` and `public/charting_library` folders must not be commited to git, please delete it.
Also, I dont see any other files than datafeeds here. Is this branch supposed to be WIP?

PR #1178, opened 2021-05-26
Feat/liquidity history 1120 

### LINKED ISSUE #1120: Liquidity page - Connect History table to the real data and add pagination
We need to connect Liquidity History table to real data and add pagination (max 6 items per page). Make it also pixel perfect and check the responsive behaviour (should be OK but just for case).

image

rick23p:
closes #1120

### DIFF
--- a/src/app/pages/LiquidityMining/components/HistoryTable/TableBody/index.tsx
+++ b/src/app/pages/LiquidityMining/components/HistoryTable/TableBody/index.tsx
@@ -1,32 +1,47 @@
+import { SkeletonRow } from 'app/components/Skeleton/SkeletonRow';
 import React from 'react';
 import { LiquidityPoolDictionary } from 'utils/dictionaries/liquidity-pool-dictionary';
 import { TableRow } from '../TableRow/index';
+import { useTranslation } from 'react-i18next';
+import { translations } from '../../../../../../locales/i18n';
+import { weiToFixed } from 'utils/blockchain/math-helpers';
 
-export const TableBody: React.FC = () => {
+interface ITableBodyProps {
+  items: any[];
+  loading: Boolean;
+}
+
+export const TableBody: React.FC<ITableBodyProps> = ({ items, loading }) => {
+  const { t } = useTranslation();
   const pools = LiquidityPoolDictionary.list();
 
   return (
     <tbody className="mt-5">
-      {pools.map((item, index) => (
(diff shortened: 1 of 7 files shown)


src/app/pages/LiquidityMining/components/HistoryTable/TableBody/index.tsx
```diff
@@ -1,32 +1,47 @@
...
+        <tr key={'loading'}>
+          <td colSpan={99}>
+            <SkeletonRow loadingText={t(translations.topUpHistory.loading)} />
+          </td>
+        </tr>
+      )}
+      {items.length === 0 && !loading && (
+        <tr key={'empty'}>
+          <td className="text-center" colSpan={99}>
+            History is empty.
```
soulBit, line 43: For this line, could you add the translation string in the same we way have for topupHistory - e.g.

`{t(translations.liquidityMining.historyTable.emptyState)}`

You will need to add this to `src/locales/en/translation.json` first, example is in object under `topUpHistory.emptyState` key

src/app/pages/LiquidityMining/components/HistoryTable/index.tsx
```diff
@@ -1,23 +1,84 @@
...
+        setHistory(events || []);
+        setTotal(pagination.totalPages * pageSize);
+        setLoading(false);
+      })
+      .catch(e => {
+        console.log(e);
+        setHistory([]);
+        setLoading(false);
+      });
+  };
```
creed-victor, line 45: I added `useFetch` hook some time ago, maybe it would be useful in cases like this

creed-victor (approved):

src/app/pages/LiquidityMining/components/HistoryTable/index.tsx
peymannaderi3: Well since it's now using new api model and pagination... useFetch needs to be changed to support this.

soulBit (approved):
Changes look good, although I noticed duplicate entries returned from backend api - Betsy will fix the issue once available

soulBit:
Duplicated tx bug has been resolved and tested now, this can be merged. I've also added a small fix for unavailable pool data (page was crashing if some pool data was missing, e.g. for new Eth/btc pool).

soulBit (approved):

creed-victor (approved):

Issue #1152, opened 2021-05-23
remove engage wallet button on spot and margin trading


t-z-u:

t-z-u:
<img width="443" alt="Screenshot 2021-05-23 at 01 26 37" src="https://user-images.githubusercontent.com/79344165/119244206-164aef00-bb66-11eb-84ef-10cc1b5f2eb2.png">

PR #1180, opened 2021-05-26
remove Engage button from spot, margin trade.

rick23p:


rick23p:
@t-z-u please check if it's working as you want

rick23p:
closes #1152

creed-victor (approved):
Approving, but in my opinion having "Engage Wallet" or any other indicator which informs user why buttons are disabled was much better approach

rick23p:
@creed-victor I agree with you. It's easier for users to know what's going on and engage easily.
@t-z-u What do you think?

pietro-maximoff (approved):

Issue #997, opened 2021-05-01
Text in *all buttons* must not wrap, implement ellipsis for long labels


bitcoinuser:
On https://redesign.sovryn.pages.dev/trade

The text Conectar Carteira should be on the same line

![Captura de tela de 2021-05-01 14-12-08](https://user-images.githubusercontent.com/1178856/116789895-b5108e00-aa87-11eb-8661-176b7374a78b.png)

bitcoinuser:
Hi, one more example on PT-BR

![Captura de tela de 2021-05-14 11-23-26](https://user-images.githubusercontent.com/1178856/118284606-fd2ba980-b4a6-11eb-90fb-4f4f8ca5b045.png)

PR #1194, opened 2021-05-27
fix buttons #997

pietro-maximoff:


tiltom (approved):

tiltom:
The changes themselves are good but we have some other buttons as well (i.e. `ActionButton` that is used on Liquidity mining page) that needs adjustment.

image

tiltom (changes requested):
The changes themselves are good but we have some other buttons as well (i.e. `ActionButton` that is used on Liquidity mining page) that needs adjustment.

image

boltsr (commented):
LGTM!

creed-victor (approved):

PR #1239, opened 2021-06-02
Arbitrage opportunities for swaps

creed-victor:


src/app/components/Arbitrage/Arbitrage.tsx
```diff
@@ -1,77 +1,120 @@
...
+import { selectWalletProvider } from '../../containers/WalletProvider/selectors';
+import { fixNumber } from '../../../utils/helpers';
+import { AssetSymbolRenderer } from '../AssetSymbolRenderer';
+import { toNumberFormat } from '../../../utils/display-text/format';
 
 const s = translations.swapTradeForm;
 
+const minUsdForOpportunity = 10;
+
+interface PoolData {
```
tiltom, line 20: I would probably create `model.ts` in the `Arbitrage` folder and move there this interface and the interface for `opportunities` so this file is less cluttered.

src/app/components/Arbitrage/Arbitrage.tsx
```diff
@@ -1,77 +1,120 @@
...
-  });
+  const { assetRates } = useSelector(selectWalletProvider);
 
-  useEffect(() => {
-    axios
-      .get(api + '/amm/arbitrage')
-      .then(res => setData(res.data))
-      .catch(console.error);
-  }, [api]);
+  const { value: data } = useFetch<PoolData>(
```
tiltom, line 38: I am getting CORS errors on local so I couldn't test it with real data but we can test it on testnet in the worst-case scenario.

src/app/components/Arbitrage/Arbitrage.tsx
```diff
@@ -1,77 +1,120 @@
...
+  const opportunityArray = useMemo(() => {
+    const opportunities: {
+      fromToken: Asset;
+      toToken: Asset;
+      fromAmount: number;
+      toAmount: number;
+      earn: number;
+      earnUsd: number;
+    }[] = [];
+    for (let pool in data) {
```
tiltom, line 52: I would maybe use `map` instead of `for`.

src/app/components/Arbitrage/Arbitrage.tsx
```diff
@@ -1,77 +1,120 @@
...
+    const items = opportunityArray.filter(
+      item => item.earnUsd > minUsdForOpportunity,
+    );
+    return items.length ? items[0] : null;
+  }, [opportunityArray]);
 
   return (
     <>
-      {show && (
+      {opportunity !== null && (
```
tiltom, line 90: The `render` needs to be adjusted for mobile devices, this is what happens under `992px` width:
image

tiltom (changes requested):

src/app/components/Arbitrage/Arbitrage.tsx
creed-victor: Oh, still looks cool :D

src/app/components/Arbitrage/Arbitrage.tsx
creed-victor: Maybe backend was down or dev branch uses wrong backend in the classifiers (should be https://api.test.sovryn.app)

tiltom (approved):

src/app/components/Arbitrage/Arbitrage.tsx
tiltom: Probably the backend was down, I tested it today and it worked like a charm.

PR #1324, opened 2021-06-12
Bug/1205-transak-fiat-receiver

boltsr:
fix transack issue


grinry (changes requested):
You need to make sure that `actions.openFiatOnRamp` works same way as 
https://github.com/DistributedCollective/Sovryn-frontend/blob/6b909938dc661d81b4390631657728f8485e87f1/src/app/containers/FastBtcDialog/saga.ts#L45
to generate address and only then open transak widget.

With current changes you only open widget and it has no deposit address - so not usable for user.

src/app/containers/FastBtcDialog/slice.ts
```diff
@@ -60,6 +60,7 @@ const fastBtcDialogSlice = createSlice({
     },
     openFiatOnRamp(state) {
       state.step = Step.TRANSAK;
+      state.deposit.loading = true;
     },
```
grinry, line 63: I think it will not be used anymore, can be deleted

grinry (changes requested):

creed-victor (approved):

PR #1356, opened 2021-06-17
[Header] Add Bridge menu option, spacing tweaks for smaller screens

soulBit:


src/app/components/Header/index.tsx
```diff
@@ -157,6 +158,13 @@ export function Header() {
     },
     { to: '/stake', title: t(translations.mainMenu.staking) },
     { to: '/wallet', title: t(translations.mainMenu.wallet) },
+    {
+      to:
+        currentNetwork === 'mainnet'
```
tiltom, line 163: I would extract this logic to a variable so you don't have to duplicate it (something like `bridgeLink`).

src/app/components/Header/index.tsx
```diff
@@ -309,7 +317,7 @@ export function Header() {
                     isSectionOpen(SECTION_TYPE.TRADE) && 'tw-font-bold'
                   }`}
                 >
-                  <span className="tw-mr-3 tw-cursor-pointer">
+                  <span className="tw-mr-1 2xl:tw-mr-3 tw-cursor-pointer">
```
tiltom, line 320: It looks ok on smaller screens but on 1200px it's too crowded. Have you discussed it with @t-z-u ? Because we wanted to have less first level items and we just added another one.

tiltom (changes requested):

PR #1358, opened 2021-06-17
Fix/1355 swap asset logic

soulBit:


src/app/components/AssetSymbolRenderer/index.tsx
```diff
@@ -30,7 +30,7 @@ export function getAssetSymbol(asset: Asset) {
   if (symbolMap.hasOwnProperty(asset)) {
     return symbolMap[asset];
   }
-  return AssetsDictionary.get(asset).symbol;
+  return AssetsDictionary.get(asset)?.symbol;
```
creed-victor, line 33: Maybe fallback to asset type itself in case it's not yet defined in dictionary?

```suggestion
  return AssetsDictionary.get(asset)?.symbol || asset;
```

creed-victor (approved):

t-z-u (approved):

tiltom (dismissed):

PR #1363, opened 2021-06-18
Remove SOV from the 'asset to collateralize' dropdown in Lend -> Borrow

bitnym:
Before this patch:
before
After this patch:
after


soulBit:
Closing as dupe of #1362

PR #1387, opened 2021-06-24
Uses swaps external smart contract for swaps.

### LINKED ISSUE #1342: SwapsExternal.sol call so stakers earn on swaps

creed-victor:
Closes #1342

Changed in:
- Swaps
- Spot trading
- Buy SOV page
- Closing opened margin trading position (check for difference between oracle and swap price, prevents position to be closed if diff > 5%)


creed-victor:
Ok, so UI should be ready however it looks like smart-contracts are failing:
1. all pairs except for the rusdt <-> doc fails
2. rusdt <-> doc tx succeeds, but instead of receiving tokens to my (receiver) wallet, tokens are sent to the protocol (for stakers maybe)

Stan is working and debugging on these issues.

Also need to find out what is difference between `requiredDestTokenAmount` and `minReturn` here, i suspect its same thing (Stan is checking on it)
https://github.com/DistributedCollective/Sovryn-frontend/pull/1387/files#diff-04105901e90b5382f624b2b719920c52f72fa2e1122aed6a83bee37cff200d08R12-R13

creed-victor:
If smart-contracts will change, most likely you will only need to change ABI file contents (SwapsExternalAbi.json), it uses sovrynProtocol address so it shouldn't change (even you receive SwapsExternal contract address you need to call swap functions on sovrynProtocol contract)

soulBit (approved):

PR #1399, opened 2021-06-29
bug/1391-margin-trade-fee

boltsr:
fix margin-trade fee issue

### DIFF
--- a/src/app/pages/MarginTradePage/components/TxFeeCalculator/index.tsx
+++ b/src/app/pages/MarginTradePage/components/TxFeeCalculator/index.tsx
@@ -36,7 +36,7 @@ export function TxFeeCalculator(props: Props) {
   );
   const gasData = React.useMemo(() => {
     const data = props.txConfig?.gas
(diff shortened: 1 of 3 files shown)


src/app/pages/MarginTradePage/components/TxFeeCalculator/index.tsx
```diff
@@ -49,7 +50,10 @@ export function TxFeeCalculator(props: Props) {
               loading={loading}
               tooltip={
                 <>
-                  {fromWei(value)} {props.symbol}
+                  {props.txConfig?.gas
+                    ? Number(props.txConfig?.gas) * Number(gas.get())
+                    : fromWei(value)}{' '}
```
creed-victor, line 55: Would like to see calculations inside useMemo and outside render, also use bignumber for math related stuff - these numbers may be very big and FE may crash because of it

creed-victor (changes requested):

creed-victor (approved):

creed-victor (approved):

PR #1417, opened 2021-07-02
Disable repay button during transactions

tiltom:


tiltom:
I am closing this PR because I will create another one with master as a base.

PR #1431, opened 2021-07-06
add maintenance to lend v2.

rick23p:


soulBit (approved):

src/app/pages/LendingPage/components/CurrencyContainer/CurrencyRow.tsx
```diff
@@ -38,13 +45,14 @@ const CurrencyRow: React.FC<Props> = ({ lendingPool, lendingAmount }) => {
           onClick={() => setDialog('add')}
           className="tw-block tw-w-full tw-mb-3 tw-rounded-lg tw-bg-ctaHover hover:tw-opacity-75"
           textClassName="tw-text-base"
+          disabled={depositLocked}
         />
         <ActionButton
```
soulBit, line 50: If `depositLocked` is set to true, can we show a Tooltip with a disabled ActionButton inside it, with tooltip text from `t(translations.maintenance.depositLend)`.

Same thing for `withdrawLocked`, but using `t(translations.maintenance.withdrawLend)` instead.

Example of this in `src\app\containers\StakePage\index.tsx::358-377` (it's using a standard `button` element, we should use a disabled `ActionButton` instead)

soulBit (changes requested):

soulBit (approved):

Issue #1425, opened 2021-07-05
Update Stake History with a new endpoints


pietro-maximoff:
/events/stake/ {{ user }} ?page=1&pageSize=1
*Currently doesn't include the FeeWithdraw event but will add

PR #1432, opened 2021-07-06
[Staking] - #1425 update staking history with a new endpoints

pietro-maximoff:


soulBit (approved):

src/app/containers/StakePage/components/HistoryEventsTable.tsx
```diff
@@ -51,80 +47,18 @@ export function HistoryEventsTable() {
...
-        .then(res => {
-          setEventsHistory(
-            (res as any).sort(
-              (x, y) =>
-                new Date(y.eventDate).getTime() -
-                new Date(x.eventDate).getTime(),
-            ),
-          );
+      axios
+        .get(backendUrl[chainId] + '/events/stake/' + account + '')
```
tiltom, line 51: You don't need `+ ''`, it's fully functional without it.

Also, you can use string interpolation and rewrite it to `${backendUrl[chainId]}/events/stake/${account}`, I think it's better for readability.

src/app/containers/StakePage/components/HistoryEventsTable.tsx
```diff
@@ -51,80 +47,18 @@ export function HistoryEventsTable() {
 
   useEffect(() => {
     async function getHistory() {
```
tiltom, line 49: Since this function is only used in one `if` and nowhere else, you don't need it to be named and you can just use its body in the `if`. Also, it doesn't need to be async because you don't `await` anything there, you use `then`.

src/app/containers/StakePage/components/HistoryEventsTable.tsx
```diff
@@ -180,25 +117,19 @@ export function HistoryEventsTable() {
...
-                        <td
-                          colSpan={5}
-                          className="tw-text-center tw-font-normal"
+                  {eventsHistory.length === 0 && (
+                    <tr>
+                      <td colSpan={5} className="tw-text-center tw-font-normal">
+                        <button
+                          type="button"
+                          className="tw-text-gold tw-tracking-normal hover:tw-text-gold hover:tw-no-underline hover:tw-bg-gold hover:tw-bg-opacity-30 tw-mr-1 xl:tw-mr-7 tw-px-4 tw-py-2 tw-bordered tw-transition tw-duration-500 tw-ease-in-out tw-rounded-full tw-border tw-border-gold tw-text-sm tw-font-light tw-font-montserrat"
+                          onClick={() => setViewHistory(true)}
```
tiltom, line 126: Create a function with `useCallback` hook and just use the function here.

src/app/containers/StakePage/components/HistoryEventsTable.tsx
```diff
@@ -222,35 +153,42 @@ interface HisoryAsset {
   index: number;
 }
 
-const HisoryTableAsset: React.FC<HisoryAsset> = ({ item, index }) => {
+const HisoryTableAsset: React.FC<HisoryAsset> = ({ item }) => {
```
tiltom, line 156: You have a typo here, it should be `HistoryTableAsset` and `HistoryAsset`.

src/app/containers/StakePage/components/HistoryEventsTable.tsx
```diff
@@ -222,35 +153,42 @@ interface HisoryAsset {
...
   const { t } = useTranslation();
   const SOV = AssetsDictionary.get(Asset.SOV);
   const dollars = useCachedAssetPrice(Asset.SOV, Asset.USDT);
   const dollarValue = useMemo(() => {
-    if (item.returnVal.amount === undefined) return '';
-    return bignumber(item.returnVal.amount)
+    if (item.amount === undefined || null) return '';
```
tiltom, line 161: Rewrite it to `if(!item.amount)` and don't write one-line `if`.

src/app/containers/StakePage/components/HistoryEventsTable.tsx
```diff
@@ -222,35 +153,42 @@ interface HisoryAsset {
...
-        <LoadableValue
-          value={numberToUSD(Number(weiToFixed(dollarValue, 4)), 4)}
-          loading={dollars.loading}
-        />
+        {item.action !== t(translations.stake.actions.delegate) ? (
+          <>
+            {numberFromWei(item.amount)} SOV
+            <br />≈{' '}
+            <LoadableValue
+              value={numberToUSD(Number(weiToFixed(dollarValue, 4)), 4)}
```
tiltom, line 181: Use our `weiTo4` method instead of `weiToFixed(..., 4)`.

src/app/containers/StakePage/components/HistoryEventsTable.tsx
```diff
@@ -299,7 +237,7 @@ const HistoryTable: React.FC<History> = ({ items }) => {
       {items.map((item, index) => {
         return (
           <HisoryTableAsset
-            key={item.transactionHash}
+            key={item.txHash + item.newBalance}
```
tiltom, line 240: `txHash` should be unique so I don't think you need to interpolate the `key`, do you?

tiltom (changes requested):

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: yes, they should be unique, but we have duplicated sometimes, need to check with @BetsyBraddock why is so
https://prnt.sc/1apo2ke

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: fixed in #1438

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: +

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: yeah, fixed

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: fixed

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: fixed

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: fixed

src/app/containers/StakePage/components/HistoryEventsTable.tsx
tiltom: It's weird but the reason can be we can have same tx hash for different action. But it's better to check with Betsy so we know we have valid data.

src/app/containers/StakePage/components/HistoryEventsTable.tsx
```diff
@@ -15,32 +16,19 @@ import { AssetsDictionary } from 'utils/dictionaries/assets-dictionary';
...
+  const { chainId } = useSelector(selectWalletProvider);
   const [eventsHistory, setEventsHistory] = useState<any>([]);
-  const [eventsHistoryVesting, setEventsHistoryVesting] = useState<any>([]);
-  const [eventsHistoryVestingTeam, setEventsHistoryVestingTeam] = useState<any>(
-    [],
-  );
-  const [eventsHistoryVestingOrigin, setEventsHistoryVestingOrigin] = useState<
-    any
-  >([]);
   const [viewHistory, setViewHistory] = useState(false);
```
tiltom, line 32: I am thinking that the naming here is a bit incorrect - we use `viewHistory` as an indicator if the history is loading. Therefore, I'd suggest renaming to `isHistoryLoading`.

tiltom (changes requested):

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: does make sense

src/app/containers/StakePage/components/HistoryEventsTable.tsx
pietro-maximoff: checked with Betsy, its a bug, will be fixed on backed side today or tomorrow, so then i'll fix it here

JoeStorm99 (dismissed):

PR #1440, opened 2021-07-08
bug/1403-history-yield

boltsr:
update message if the wallet disconnect

### DIFF
--- a/src/app/components/FinanceV2Components/CardRow/index.tsx
+++ b/src/app/components/FinanceV2Components/CardRow/index.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+
 import { StyledCardRow } from './styled';
 
 interface ICardRowProps {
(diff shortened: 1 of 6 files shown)


soulBit (changes requested):
@boltsr your changes look good, but I spotted one small issue with styling on the skeleton component, it's breaking out of the container on horizontal axis and causing scrollbar to appear, due to negative margin.
In `Sovryn-frontend\src\app\components\Skeleton\SkeletonRow.tsx`, remove `tw--mx-4` from lines 10 and 21

boltsr:
updated @soulBit

src/app/components/FinanceV2Components/CardRow/index.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+
 import { StyledCardRow } from './styled';
 
 interface ICardRowProps {
```
tiltom, line 5: I would adjust the interface to accept `className` and use it in the top div. This way, we could adjust the component and add `tw-mt-2` in both of the usages in this story because it looks a bit weird without a top margin (we use `tw-mt-2` in Portfolio page as well).

tiltom (changes requested):

Issue #1438, opened 2021-07-07
investigate why 2 different staked amounts are showing


t-z-u:

t-z-u:
<img width="489" alt="Screenshot 2021-07-07 at 12 43 45" src="https://user-images.githubusercontent.com/79344165/124753658-2a358f00-df21-11eb-8ebf-dfc1d9f51738.png">

because i have only one stake the "total staked sov" and "locked amount" should be the same number

PR #1441, opened 2021-07-09
[Staking] - #1438 fixed rounding issue

pietro-maximoff:


src/app/containers/StakePage/components/CurrentStakes.tsx
```diff
@@ -195,7 +194,7 @@ function AssetRow(props: AssetProps) {
         </div>
       </td>
       <td className="tw-text-left tw-font-normal">
-        {weiToNumberFormat(props.item[0])} {t(translations.stake.sov)}
+        {numberFromWei(props.item[0])} {t(translations.stake.sov)}
```
tiltom, line 197: I don't think we need this level of precision. I would use `weiToFixed(props.item[0], 3)`since we display only 3 decimals in `Total Staked SOV`.

image

src/app/containers/StakePage/components/VestingContract.tsx
```diff
@@ -168,7 +168,7 @@ export function VestingContract(props: Props) {
               <p className={`tw-m-0 ${lockedAmount.loading && 'skeleton'}`}>
                 {lockedAmount.value && (
                   <>
-                    {weiToNumberFormat(lockedAmount.value)}{' '}
+                    {numberFromWei(lockedAmount.value)}{' '}
```
tiltom, line 171: The same, I would use `weiToFixed(lockedAmount.value, 3)`.

tiltom (changes requested):

pietro-maximoff:
discussed with Tzu, so the best view is 0.0000

src/app/containers/StakePage/components/CurrentStakes.tsx
```diff
@@ -194,15 +194,15 @@ function AssetRow(props: AssetProps) {
         </div>
       </td>
       <td className="tw-text-left tw-font-normal">
-        {numberFromWei(props.item[0])} {t(translations.stake.sov)}
+        {weiToFixed(props.item[0], 4)} {t(translations.stake.sov)}
```
tiltom, line 197: It looks good now, just use `weiTo4` function for every file in this PR.

tiltom (changes requested):

src/app/containers/StakePage/components/CurrentStakes.tsx
pietro-maximoff: yeah, its done

soulBit (approved):

PR #1487, opened 2021-07-15
Allow yield farm interactions when a wallet is not connected

tiltom:


soulBit (dismissed):

bitcoinuser:
In my opinion the button Deposit should be disabled when wallet is not connected.

JoeStorm99 (approved):

PR #1497, opened 2021-07-16
Align buttons in margin trade dialog

tiltom:


tiltom:
I added truncation to the buttons but in order to not have it truncated in English, I needed to make dialogs a bit wider (340px instead of 320px). I think we should redesign buttons because the text is too large and it's often truncated in Portuguese (and it could be the case in other languages as well).

rick23p (approved):

JoeStorm99 (approved):

soulBit (approved):

creed-victor:
> I added truncation to the buttons but in order to not have it truncated in English, I needed to make dialogs a bit wider (340px instead of 320px). I think we should redesign buttons because the text is too large and it's often truncated in Portuguese (and it could be the case in other languages as well).

Most smaller mobile devices have screen width of 320px, 340px will show horizontal scroll for such devices. We "don't support" mobile devices, but it's something to consider for when we do.

tiltom:
> > I added truncation to the buttons but in order to not have it truncated in English, I needed to make dialogs a bit wider (340px instead of 320px). I think we should redesign buttons because the text is too large and it's often truncated in Portuguese (and it could be the case in other languages as well).
> 
> Most smaller mobile devices have screen width of 320px, 340px will show horizontal scroll for such devices. We "don't support" mobile devices, but it's something to consider for when we do.

It's always a tradeoff but I think that large majority of our users will more appreciate not to have it truncated on all of their devices. What do you think @soulBit or @t-z-u ?

creed-victor (dismissed):

creed-victor (approved):

PR #1505, opened 2021-07-18
[Lend Page] XUSD profit display fix

soulBit:


src/app/pages/LendingPage/components/CurrencyContainer/UserLendingInfo.tsx
```diff
@@ -47,11 +50,36 @@ export const UserLendingInfo: React.FC<IUserLendingInfoProps> = ({
     account,
   );
 
+  const {
+    value: balanceOfCall,
```
tiltom, line 54:
```suggestion
    value: totalBalance,
```

src/app/pages/LendingPage/components/CurrencyContainer/UserLendingInfo.tsx
```diff
@@ -47,11 +50,36 @@ export const UserLendingInfo: React.FC<IUserLendingInfoProps> = ({
...
   const balance = useMemo(() => {
     return bignumber(balanceCall).minus(profitCall).toString();
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [balanceCall, profitCall, asset]);
 
+  const totalProfit = useMemo(() => {
+    return bignumber(tokenPrice)
+      .sub(checkpointPrice)
+      .mul(balanceOfCall)
+      .div(10e18)
```
tiltom, line 77: I don't like magic numbers in a code but I suppose this particular case should be fine for now. However, we would have problems if we add a token with more (or less) than 18 decimals.

tiltom (changes requested):

src/app/pages/LendingPage/components/CurrencyContainer/UserLendingInfo.tsx
```diff
@@ -47,11 +50,36 @@ export const UserLendingInfo: React.FC<IUserLendingInfoProps> = ({
...
+      .add(profitCall)
+      .toString();
```
creed-victor, line 79: If there is a chance that wei amount will have a decimals - make sure to remove them, otherwise dapp may crash because wei can be integers only.

```suggestion
      .toFixed(0);
```

tiltom (approved):

creed-victor (approved):

PR #1512, opened 2021-07-19
Disable swap button when wallet is disconnected

tiltom:


creed-victor (approved):

src/app/containers/SwapFormContainer/index.tsx
```diff
@@ -321,7 +321,12 @@ export function SwapFormContainer() {
...
         )}
         <BuyButton
-          disabled={tx.loading || (!validate && isConnected) || swapLocked}
+          disabled={
+            tx.loading ||
+            !isConnected ||
+            (!validate && isConnected) ||
+            swapLocked
+          }
           onClick={() => onSwap()}
```
creed-victor, line 330: `onSwap` looks like this:

````
  const onSwap = () => {
    if (isConnected) send();
    else connect();
  };
```

So it means that `onSwap` body is wrong or disabling button when user is not connected is wrong.

No idea which approach is correct, @t-z-u?

In my opinion button should be enabled but label conditional (show "Egage Wallet" when not connected)

soulBit: Current behaviour is correct as per Tzu's most recent request to update this, I have made the logic consistent now - no need for isConnected check since button is already disabled in this state

PR #1532, opened 2021-07-21
Staging deployment workflow

grinry:
- Github workflow to deploy any selected branch as staging environment (https://staging.sovryn.app) using github actions (manual deployment only, no PR/commit triggers - thats intentional). 
- Moved charting-library dependency from private repository to private github package - developers will be required to login npm (for github registry) or create github token - see updated README


creed-victor:
Deploying to AWS finished, however deploying to CloudFlare Pages will not work anymore until this is fixed: https://github.com/cloudflare/cloudflare-docs/issues/1820

ShynRou:
Putting the access token definition in the .npmrc is bad, since it's a tracked file it's inevitable that someone will commit it with their token. Would be better to place it there with a env variable. And let the user set it on their machine. Also the .env.local files shouldn't be tracked for the same reason.

```
//npm.pkg.github.com/:_authToken=$GITHUB_NPM_PKG_TOKEN
```

ShynRou (approved):

.github/workflows/deploy-staging.yml
```diff
@@ -6,41 +6,44 @@ jobs:
...
-          #            invalidation: /
-          delete-removed: true
-          no-cache: false
-          private: true
+        uses: raulanatol/aws-s3-docker-action@master
+        env:
+          AWS_ACCESS_KEY_ID: '${{ secrets.AWS_ACCESS_KEY_ID }}'
+          AWS_SECRET_ACCESS_KEY: '${{ secrets.AWS_SECRET_ACCESS_KEY }}'
+          AWS_BUCKET_NAME:  '${{ secrets.AWS_S3_BUCKET }}'
+          AWS_REGION: us-east-2
```
tiltom, line 43: Why did you choose the `us-east-2` region? We are mostly located in Europe so I think it would make more sense to use some EU datacenter because of the latency.

creed-victor: This is where devops puts buckets :D Also, everything is also cached and distributed across cloudflare datacenters

tiltom: Ok, you're probably right that cloudfare mitigates the problem :D

tiltom (approved):

PR #1569, opened 2021-07-27
Feat/1494 replace moment

### LINKED ISSUE #1494: Replace moment.js
moment.js is to heavy and requires a hack in the webpack config to work properly, should be replaced with smaller simpler sollution.

- [ ] replace moment.js
- [ ] replace moment-timezones.js
- [ ] remove IgnorePlugin in webpack config

ShynRou:
Resolves #1494

Replaced moment with dayjs which is lighter, supports localized formatting and does not require a webpack hack for treeshaking.


src/app/components/StakingDateSelector/index.tsx
```diff
@@ -44,21 +44,24 @@ export function StakingDateSelector(props: Props) {
...
 
-  const dateWithoutStake = filteredDates.reduce((unique: any, o: any) => {
-    let isFound = itemDisabled.some((b: { key: any }) => {
-      return b.key === o.key;
-    });
-    if (!isFound) unique.push(o);
-    return unique;
-  }, []);
+  const [dateWithoutStake, avaliableYears, avaliableMonth] = useMemo(() => {
+    const dateWithoutStake = filteredDates.reduce((unique: any, o: any) => {
```
tiltom, line 48: I don't like this, use correct typing. `unique` is `DateItem[]` and `o` is `DateItem`. 

Also, adjust the naming - change `unique` to `uniqueDates` and `o` to `item` (or something else if you find it more suitable).

tiltom (changes requested):

tiltom (dismissed):

creed-victor (approved):

soulBit (approved):
Removed `antd` package dependency as discussed

PR #1584, opened 2021-08-02
Update loot drop promos date

tiltom:
https://taiga.sovryn.app/project/sovryn-light/task/67

### DIFF
--- a/src/app/pages/LendingPage/index.tsx
+++ b/src/app/pages/LendingPage/index.tsx
@@ -35,7 +35,7 @@ const LendingPage: React.FC = () => {
             title="15K SOV"
             asset1={Asset.XUSD}
             message={t(translations.liquidityMining.recalibration, {
-              date: 'August 2',
+              date: 'August 9',
             })}
             linkUrl="https://www.sovryn.app/blog/sov-is-diving-into-the-lending-pools"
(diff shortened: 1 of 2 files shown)


soulBit (approved):

rick23p (approved):

rick23p:
What do you think guys if we could make a LootDrop JSON and put all loot drops in there and read from it in different places?
@tiltom @soulBit

tiltom:
> What do you think guys if we could make a LootDrop JSON and put all loot drops in there and read from it in different places?
> @tiltom @soulBit

This could be good if we wanted non-technical people to be able to adjust the loot drops but I doubt they will want to do it. We would also need to adjust the code for it.

What do you think @soulBit ?

soulBit:
@rick23p @tiltom let's discuss this in standup today, I agree that we should find a less manual way of updating these

tiltom:
> @rick23p @tiltom let's discuss this in standup today, I agree that we should find a less manual way of updating these

Ok we can discuss it but it could work as Rick proposed - we would have a JSON file that could be edited by anyone and we would read loot drop info from it. But it needs validation with marketing people that they will use it because we would need to implement it.

PR #1589, opened 2021-08-04
[Margin] - added loanEvents endpoint for closed margin trading positi…

pietro-maximoff:
…ons P/L calculation

- added //eslint-disable-next-line to trading chart to prevent compile warnings


src/app/pages/MarginTradePage/components/TradingHistory/index.tsx
```diff
@@ -208,16 +211,22 @@ export function TradingHistory() {
...
+            .then(response => {
+              return response.json();
+            })
+            .then(loanEvents => {
+              const calculation = calculateProfits(events, loanEvents);
+              if (calculation) {
+                closeEntries.push(calculation);
+              }
+            })
+            .catch(console.error);
```
creed-victor, line 227: I have hundreds of items in my wallet and takes a minute to fetch data for each of loan.
I think fetching should be removed from here and moved to history row component - so it would fetch only when when component is actually rendered.
So entire row for user would be visible instantly and only profit calculation would loan a little bit

creed-victor (changes requested):

creed-victor (approved):

src/app/components/TradeProfit/index.tsx
```diff
@@ -1,70 +1,114 @@
...
+              .toNumber();
+          }
+          setProfit(
+            bignumber(change).mul(bignumber(positionSize)).div(100).toFixed(0),
+          );
+          setProfitDirection(change);
+        }
+      });
+    })
+    .catch(console.error);
```
creed-victor, line 72: Put it to `useEffect` to make sure it's being called only once, cancel request when component unmounts.

creed-victor (changes requested):

pietro-maximoff:
closed without merging, moved changes to this branch - feat/281-new-margin-trade-page-base-branch

PR #1590, opened 2021-08-06
fix Arbitrage list issue.

rick23p:


src/utils/helpers.ts
```diff
@@ -206,3 +206,7 @@ export const abbreviateNumber = (
   }
   return shortValue + suffixes[suffixNum];
 };
+
+export const isEmpty = val => {
+  return val === undefined || val == null ? true : false;
```
tiltom, line 211:
```suggestion
   val === undefined || val == null;
```

src/utils/helpers.ts
```diff
@@ -206,3 +206,7 @@ export const abbreviateNumber = (
   }
   return shortValue + suffixes[suffixNum];
 };
+
+export const isEmpty = val => {
```
tiltom, line 210:
```suggestion
export const isNullOrUndefined = val => {
```

src/app/components/Arbitrage/Arbitrage.tsx
```diff
@@ -31,7 +31,16 @@ export function Arbitrage() {
   const opportunityArray = useMemo(
     () =>
       Object.values(data)
-        .filter(item => item.hasOwnProperty('rateToBalance'))
+        .filter(item => {
+          const rateToBalance = item?.rateToBalance;
```
tiltom, line 35: You can extract this whole check to a local function in this file (e.g. `isValidArbitrage`).

tiltom (changes requested):

src/utils/helpers.ts
```diff
@@ -207,6 +207,6 @@ export const abbreviateNumber = (
   return shortValue + suffixes[suffixNum];
 };
 
-export const isEmpty = val => {
-  return val === undefined || val == null ? true : false;
+export const isNullOrUndefined = val => {
```
tiltom, line 210: You have unnecessary curly braces and `return` here.

tiltom (changes requested):

src/utils/helpers.ts
```diff
@@ -206,3 +206,5 @@ export const abbreviateNumber = (
   }
   return shortValue + suffixes[suffixNum];
 };
+
+export const isNullOrUndefined = val => val === undefined || val == null;
```
ShynRou, line 210: `val == null` already checks for null and undefined.

In my opinion this function is redundant, you could just check `rateToBalance.earn != null`.

tiltom: Usage of `==` in JS is antipattern and I wouldn't use it at all.

tiltom: And I haven't noticed it's used here, good catch :D @rick23p please change `val == null` to `vall === null`.

creed-victor (approved):

tiltom (approved):

PR #1599, opened 2021-08-11
Show connected user wallet address in babelfish flows

creed-victor:
- shows receiver wallet (wallet originally connected to rsk chain) in deposit flow
- shows currently connected (rsk) wallet in withdrawal flow
- changes "available balance" with "max withdrawal" in withdraw flow

### DIFF
--- a/src/app/components/LinkToExplorer/index.tsx
+++ b/src/app/components/LinkToExplorer/index.tsx
@@ -1,11 +1,4 @@
-/**
- *
- * LinkToExplorer
- *
- */
-import React, { useCallback, useEffect, useState } from 'react';
-import { useSelector } from 'react-redux';
-import { selectWalletProvider } from '../../containers/WalletProvider/selectors';
+import React, { useMemo } from 'react';
 import { blockExplorers, currentChainId } from '../../../utils/classifiers';
 
 interface Props {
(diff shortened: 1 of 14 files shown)


bitcoinuser:
Hi @creed-victor. I'm testing it and I see a bug:

On Metamask usually when needed to change network, the Metamask window popup. In this case it doesn't, even clicking on BSC Network button on center.

Captura de tela de 2021-08-11 10-48-43

creed-victor:
> Hi @creed-victor. I'm testing it and I see a bug:
> 
> On Metamask usually when needed to change network, the Metamask window popup. In this case it doesn't, even clicking on BSC Network button on center.
> 

I think @rick23p is implementing this metamask feature as another task.

tiltom:
I don't like that sometimes we have information on two rows and it's unnecessary because we have a ton of space. I know it's probably not so relevant to the PR but I think we can adjust it here since it would be really quick fix.

image

rick23p:
> @t-z-u , @creed-victor - I don't like that sometimes we have information on two rows and it's unnecessary because we have a ton of space. I know it's probably not so relevant to the PR but I think we can adjust it here since it would be really quick fix.
> 
> image

increased the width on https://github.com/DistributedCollective/Sovryn-frontend/pull/1606

tiltom (approved):

bitcoinuser:
UI improvement: On my screen buttons are very near. See:

Captura de tela de 2021-08-16 10-30-26

bitcoinuser:
Captura de tela de 2021-08-16 10-35-27

PR #1601, opened 2021-08-11
Updated supported browsers

ShynRou:
As discussed with @t-z-u, we only aim to support newish browser versions to not get bogged down by old or crappy browsers that have bad feature support. To represent that I updated the browserslist to support up to roughly 2 year old browser versions. 

Note: Brave is currently not supported by browserslist, but the chrome version mostly covers it.


t-z-u:
i only use brave!!!

t-z-u (changes requested):
pls make brave work in the list

ShynRou:
@t-z-u as explained in discord, brave is supported by us just not by the tool "browserslist", that we use for automatically handling browser feature detection/handling. But since Brave is based on chromium and only differs in few features, we should be fine with having chrome in the list. 

https://github.com/browserslist/browserslist/issues/578
https://github.com/Fyrd/caniuse/issues/2749

creed-victor (approved):

soulBit (approved):

PR #1606, opened 2021-08-16
Fix/cross chain issue

rick23p:


creed-victor (approved):

src/app/pages/BridgeDepositPage/components/WalletSelector/index.tsx
```diff
@@ -64,6 +70,26 @@ export function WalletSelector() {
     chain,
   ]);
 
+  const changeNetwork = () => {
```
tiltom, line 73: Use `useCallback`. 

Also, I would check at the beginning if the network exists and if not, return. If it does exist, destructure the network object so you don't have to write e.g. `network?.explorer` but `explorer` instead.

src/utils/metamaskHelpers.ts
```diff
@@ -1,4 +1,6 @@
-const addNetwork = (params: any) => {
+export const metamaskDefaultChains = [1, 3, 4, 5, 42];
+
+export const addNetwork = (params: any) => {
   const { ethereum } = window as any;
```
tiltom, line 4: You could export `ethereum` the same way as `metamaskDefaultChains` so you don't have to redeclare it in every method.

tiltom (changes requested):

tiltom (approved):

PR #1610, opened 2021-08-16
Fix types in SwapAssetSelector

tiltom:


src/app/components/Form/AssetSelect/index.tsx
```diff
@@ -26,10 +26,8 @@ export function AssetSelect(props: Props) {
   return (
     <Select
       value={props.value as any}
```
ShynRou, line 28: Select should accept a nullish value, no casting should be required.

ShynRou (changes requested):

src/app/components/Form/AssetSelect/index.tsx
tiltom: I wanted to change it but unfortunately, we don't have any current usage of this. However, I can change it and we can adjust the usage later.

ShynRou: Okey.

ShynRou (approved):

src/app/components/Form/AssetSelect/index.tsx
tiltom: Changed + `Select` was adjusted as well.

ShynRou (approved):

creed-victor (approved):

PR #1613, opened 2021-08-17
[LendingPage] fix HistoryTable timestamp type mismatch

ShynRou:
Resolves https://taiga.sovryn.app/project/sovryn-light/issue/278


rick23p (dismissed):

tiltom (dismissed):
The correct fix is changing `ITableRowProps` because `item.time` is number but it's currently incorrect typed. I have this fix in https://github.com/DistributedCollective/Sovryn-frontend/pull/1611 so we can merge your PR as a quickfix and I will merge it in mine.

ShynRou:
@tiltom Just updated the LendingEvent Interface to actually match what comes back from the API.
Stuff like this is why we need a better API implementation.

tiltom (approved):

grinry (approved):

PR #1636, opened 2021-08-24
[Staking] - fixed issues with max dates

pietro-maximoff:


soulBit (approved):
Confirmed that this resolves the staking issues from taiga ticket 314, with the exception of voting power issue when staking beyond May 3rd 2024

PR #1639, opened 2021-08-24
StakePage design cleanup

ShynRou:
Extension to #1630, in which I replaced value skeletons with spinners.

I updated the stake page, removed unnecessary borders and backgrounds and adjusted the margins.

@t-z-u can you please check, if the result is up to your standards?


ShynRou:
Since Cloudflare couldn't be bothered: https://6666f837.sovryn.pages.dev

PR #1670, opened 2021-08-30
Implement Margin Trade notification settings

tiltom:


creed-victor (approved):
This shouldn't be merged until backend is ready for this? @tiltom

rick23p (approved):
@tiltom Looks good, the only thing is that we need this to be used on new spot trading page too. so we should make it a bit more reusable. maybe we can do it once the backed is ready.

tiltom:
> This shouldn't be merged until backend is ready for this? @tiltom

This could be merged because I only want to merge it to a new margin trade base branch. BE integration of notification settings will be added in https://taiga.sovryn.app/project/sovryn-light/task/282? .

PR from the base branch to development doesn't exist yet but I will create it (as a Draft) once this PR is merged and label it with `do-not-merge`.

tiltom:
> @tiltom Looks good, the only thing is that we need this to be used on new spot trading page too. so we should make it a bit more reusable. maybe we can do it once the backed is ready.

That information is new to me :D Do you know if we're going to launch both Spot and Margin trade new pages at once? If so, you could merge our base branch to yours.

grinry:
> @tiltom Looks good, the only thing is that we need this to be used on new spot trading page too. so we should make it a bit more reusable. maybe we can do it once the backed is ready.

What is the use trade on spot trade? Notify user when price drops/increases to user target? Or it's for spot trading using order books?

ShynRou (approved):

tiltom:
> > @tiltom Looks good, the only thing is that we need this to be used on new spot trading page too. so we should make it a bit more reusable. maybe we can do it once the backed is ready.
> 
> What is the use case for it on spot trade? Notify user when price drops/increases to user target? Or it's for spot trading using order books?

@rick23p ?

Issue #1666, opened 2021-08-27
On https://live.sovryn.app/origins babelfish sale some info appear as TBC


bitcoinuser:
Hi,

On https://live.sovryn.app/origins babelfish sale some info appear as TBC, like: 

Total Raised:
Participating Wallets:

PR #1673, opened 2021-08-30
fix origins items.

rick23p:


tiltom (approved):

creed-victor:
@rick23p add 1022 instead of TBC as participating wallet of finished FISH sale too (as per #1666)

creed-victor (approved):

PR #1732, opened 2021-09-15
feat: added rif token and rbtc/rif amm converter pool

creed-victor:
Testnet contracts only, do not merge until mainnet contracts are ready too!

testing:
1. verify RIF token is in user portfolio
2. verify you can trade RIF token in swap page
3. verify you can trade RIF token in spot page
4. verify you add liquidity to RBTC/RIF amm pool (yield farming page), you will need some RIF and RBTC in your wallet (buy RIF in swap or spot page)
5. verify you can remove liquidity from RBTC/RIF amm pool after you added it in step 4.


src/utils/blockchain/contracts.ts
```diff
@@ -107,6 +107,21 @@ export const contracts = {
     abi: LiquidityPoolV1Converter,
     blockNumber: 1218833,
   },
+  RIF_token: {
+    address: '0x19f64674d8a5b4e652319f5e239efd3bc969a1fe', // todo
+    abi: TestTokenABI,
```
tiltom, line 112: I know this doesn't have anything with this PR but why is the ABI called `TestTokenABI`? From what I've seen in the code, it's used for some "altcoins" that have just some common functionality. I would suggest renaming it to something more appropriate, such as `CommonTokenABI` or `GeneralTokenABI` (and rename the abi file itself as well).

tiltom (changes requested):

src/utils/blockchain/contracts.ts
creed-victor: Indeed, it should be called `ERC20TokenABI`, but will make adjustments on another PR when all other PR's with this abi as dependency are merged

tiltom (approved):

PR #1738, opened 2021-09-16
Perpetual Contract Details

ShynRou:
Initial implementation of https://taiga.sovryn.app/project/sovryn-light/task/441

### DIFF
--- a/src/app/pages/PerpetualPage/components/ContractDetails/index.tsx
+++ b/src/app/pages/PerpetualPage/components/ContractDetails/index.tsx
@@ -1,0 +1,79 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { translations } from '../../../../../locales/i18n';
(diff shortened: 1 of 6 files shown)


tiltom (changes requested):
`24hr Volume`, `Open Interest`, `Contract Value` and `Min Trade Amount` should have `USD` after values. Otherwise it looks ok.

ShynRou:
@tiltom good catch! I used the longAsset of the trading pair for it. I'm not sure if that's correct, but just USD seems (more) wrong to me.

tiltom:
> @tiltom good catch! I used the longAsset of the trading pair for it. I'm not sure if that's correct, but just USD seems (more) wrong to me.

yeah, it's definitely not ok to hardcode something so this is probably the best solution for now :) we can adjust it later if needed

tiltom (approved):

PR #1740, opened 2021-09-16
SOV margin trading (long)

creed-victor:
Testing:
1. go to margin trading page
2. select sov-xusd pair
3. verify that pair has only "LONG" button available
4. open position with SOV or XUSD as collateral, verify tx confirms
5. "add to margin" of opened sov-xusd pair, verify tx confirms
6. close position of opened sov-xusd pair, verify tx confirms
7. repeat 4-6 with another collateral (not used on step 4 already).
8. repeat 2-7 steps with pairs rbtc-sov, bpro-sov and sov-doc.

### DIFF
--- a/src/utils/dictionaries/trading-pair-dictionary.ts
+++ b/src/utils/dictionaries/trading-pair-dictionary.ts
@@ -10,9 +10,9 @@ export enum TradingPairType {
   BPRO_XUSD = 'BPRO_XUSD',
   BPRO_USDT = 'BPRO_USDT',
   BPRO_DOC = 'BPRO_DOC',
+  BPRO_SOV = 'BPRO_SOV',
   SOV_XUSD = 'SOV_XUSD',
   SOV_DOC = 'SOV_DOC',
-  BPRO_SOV = 'BPRO_SOV',
 }
 
 export class TradingPairDictionary {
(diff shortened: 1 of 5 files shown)


src/utils/dictionaries/trading-pair-dictionary.ts
```diff
@@ -11,14 +10,12 @@ export enum TradingPairType {
   BPRO_XUSD = 'BPRO_XUSD',
   BPRO_USDT = 'BPRO_USDT',
   BPRO_DOC = 'BPRO_DOC',
+  SOV_XUSD = 'SOV_XUSD',
+  SOV_DOC = 'SOV_DOC',
+  BPRO_SOV = 'BPRO_SOV',
```
tiltom, line 15: This is just a question but why do we have `SOV_XUSD`, `SOV_DOC` and then suddenly `BPRO_SOV`? Or the order doesn't matter?

tiltom (changes requested):

src/utils/dictionaries/trading-pair-dictionary.ts
creed-victor: Yep, order doesn't matter but will reorder it because of readability

tiltom: ok thanks :)

tiltom (approved):

PR #1741, opened 2021-09-17
Origin FISH vesting withdraw

creed-victor:
Testing:
1. open portfolio page -> vested assets
2. verify "Origin FISH" row is visible
3. verify "Origin FISH" has "Withdraw" button
4. Withdraw your unlocked balance if any, verify tx confirms.

"Origin FISH" row will not appear if your wallet has no FISH vestings (didn't participate in testnet sale)


src/app/components/UserAssets/VestedAssets.tsx
```diff
@@ -1,64 +1,34 @@
...
   const connected = useIsConnected();
   const account = useAccount();
 
-  const item = AssetsDictionary.get(Asset.CSOV);
-  const result = useVestedStaking_balanceOf(account);
+  const { loading, items } = useListOfUserVestings();
 
   const [open, setOpen] = useState(false);
-  const [address, setAddress] = useState('');
+  const [vesting, setVesting] = useState<FullVesting>(null!);
```
tiltom, line 21: This seems a bit fishy :D If you want it to accept null, adjust it to `useState<Nullable<FullVesting>>(null)` because `!` will just tell Typescript to shut up but it's not correct.

src/app/components/UserAssets/VestedAssets.tsx
```diff
@@ -1,64 +1,34 @@
...
-      result.babelFishVestedValue !== '0',
-    [
-      result.babelFishVestedValue,
-      result.lmVestingContract,
-      result.originVestedValue,
-      result.teamVestedValue,
-      result.vestedValue,
-    ],
-  );
+    setVesting(null!);
```
tiltom, line 30: The same hack as on line 21 (but it won't be necessary if you adjust it as I said).

src/app/components/UserAssets/VestedAssets.tsx
```diff
@@ -99,60 +69,16 @@ export function VestedAssets() {
             )}
             {connected &&
```
tiltom, line 70: I would just extract this condition outside the render to some variable (e.g. `isVestingLoaded`) and use `useMemo`.

src/app/components/UserAssets/Vesting/VestedItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const VestedItem: React.FC<VestedItemProps> = ({ vesting, onWithdraw }) => {
+  const { logoSvg, symbol, decimals } = AssetsDictionary.get(vesting.asset);
+
+  const { t } = useTranslation();
+  const { checkMaintenance, States } = useMaintenance();
+  const withdrawLocked = checkMaintenance(States.WITHDRAW_VESTS);
+  const dollars = useCachedAssetPrice(vesting.asset, Asset.USDT);
+  const dollarValue = useMemo(() => {
+    if ([Asset.USDT, Asset.DOC].includes(vesting.asset)) {
```
tiltom, line 32: What about `Asset.XUSD`? I think it should be included in this list.

src/app/components/UserAssets/Vesting/VestedItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+      return vesting.balance;
+    } else {
+      return bignumber(vesting.balance)
+        .mul(dollars.value)
+        .div(10 ** decimals)
+        .toFixed(0);
+    }
+  }, [dollars.value, vesting.balance, vesting.asset, decimals]);
+
+  const handleOnWithdraw = useCallback(() => {
```
tiltom, line 42: Remove unnecessary curly braces.

src/app/components/UserAssets/Vesting/VestedItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+          className="tw-inline tw-mr-2 tw-h-8 tw-w-8"
+          src={logoSvg}
+          alt={vesting.asset}
+        />{' '}
+        {vesting.label || symbol}
+      </td>
+      <td className="tw-text-right">{weiToNumberFormat(vesting.balance, 4)}</td>
+      <td className="tw-text-right">
+        <LoadableValue
+          value={numberToUSD(Number(weiToFixed(dollarValue, 4)), 4)}
```
tiltom, line 59:
```suggestion
          value={numberToUSD(Number(weiTo4(dollarValue)), 4)}
```

src/app/components/UserAssets/Vesting/VestedItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+            text={t(translations.userAssets.actions.withdraw)}
+            onClick={handleOnWithdraw}
+          />
+        )}
+      </td>
+    </tr>
+  );
+};
+
+export default VestedItem;
```
tiltom, line 89: Why do you use default export here? I think you can use named.

src/app/components/UserAssets/Vesting/VestingWithdrawForm.tsx
```diff
@@ -0,0 +1,147 @@
...
+            })}
+          >
+            <InputField
+              onChange={event => setAddress(event.target.value)}
+              value={address}
+            />
+          </FieldGroup>
+
+          <div className={styles.txFee}>
+            {t(translations.common.fee, { amount: '0.000014' })}
```
tiltom, line 116: Do you want to hardcode the fee?

tiltom (changes requested):

src/app/components/UserAssets/Vesting/VestedItem.tsx
grinry: USD rate is actually RUSDT, and doc is aleays == RUSDT, but XUSD is part of amm so it may be slightly different :)

src/app/components/UserAssets/Vesting/VestingWithdrawForm.tsx
grinry: Yes, due to the bug on rsk node this call will almost always estimate far less gas than required sk value is hardcoded to the maximum function would use

src/app/components/UserAssets/Vesting/VestedItem.tsx
tiltom: ok, I didn't know that :)

src/app/components/UserAssets/Vesting/VestingWithdrawForm.tsx
tiltom: ok, thanks for the explanation

src/app/components/UserAssets/index.tsx
```diff
@@ -254,8 +244,8 @@ function AssetRow({
       </td>
       <td className="tw-text-right tw-hidden md:tw-table-cell">
         <LoadableValue
-          value={numberToUSD(Number(weiTo4(dollarValue)), 4)}
-          loading={dollars.loading}
+          value={numberToUSD(Number(weiTo4(dollarValue.value)), 4)}
```
tiltom, line 247: Everything looks good now but I found all of the references of `numberToUSD` and I found out all of them were used with 4 decimals. Could you adjust `numberToUSD` function to have 4 as a default decimal value and then you can use it without the parameter? Also, please adjust all of the places where it's used so we don't have any new occurrences of it.

tiltom (changes requested):

src/app/components/UserAssets/index.tsx
creed-victor: Not sure all of numberToUSD are using 4, as I see in the homepage there is 2, 3 or even 0 decimal places used

tiltom: I can see this when I find all the occurrences but I think it's worth the refactoring even if we have some places where it would be different (that's why we have it parametrized) :)

image

creed-victor: Ok, refactored and made new helper `weiToUSD` so there wouldn't be need to use numberToUSD, weiToFixed and Number each time.

tiltom (approved):

PR #1742, opened 2021-09-17
Remove Bitcoin prefix from FastBTC QR code

tiltom:


tiltom:
I discussed it with @creed-victor and we will leave it as it is, for more context see the discussion here: https://discord.com/channels/729675474665603133/811577759859998762/885122378768715807

PR #1766, opened 2021-09-24
Chart adjustments

tiltom:


src/app/pages/PerpetualPage/components/DepthChart/index.tsx
```diff
@@ -99,8 +99,10 @@ export const DepthChart: React.FC<IDepthChartProps> = ({ className }) => {
       },
       yAxis: [
         {
+          tickAmount: 6,
           lineWidth: 1,
           gridLineWidth: 1,
+          gridLineColor: '#484848',
```
rick23p, line 105: isn't there any better way to not use hard coded colors (reading from a file)?

rick23p (approved):

src/app/pages/PerpetualPage/components/DepthChart/index.tsx
tiltom: I could use CSS variables, but this is just a configuration, so I cannot use it here. We already have our color palette defined in `tailwind.config.js` so I don't think that its duplication would be good. 

We can think about some other way if we see we need to hardcode colors often but I would leave it this way for now.

PR #1772, opened 2021-09-27
Staking & Vesting Updates

pietro-maximoff:


soulBit:
@pietro-maximoff did some testing on this PR, portfolio page (vested assets section) has following error:

image

Stake page throws the following error:

image

seems like contract calls need to be updated to use new method args?

src/app/components/UserAssets/Vesting/useListOfUserVestings.ts
```diff
@@ -1,121 +1,91 @@
...
+                  : '';
+              const labelReward =
+                typeCreation[index] === '3' && type[index] === '1'
+                  ? 'reward'
+                  : '';
+
+              return {
+                asset: Asset.SOV,
+                staking: 'staking',
+                type: labelGenesys || labelTeam || labelReward || labelOrigin,
```
ShynRou, line 62: Just a suggestion:

```suggestion
              const label = {
                '0-1': 'team',
                '1-1': 'genesis',
                '1-2': 'origin',
                '1-3': 'reward',
              }[`${type[index]}-${typeCreation[index]}`] || '';

              return {
                asset: Asset.SOV,
                staking: 'staking',
                type: label,
```

src/app/components/UserAssets/Vesting/useListOfUserVestings.ts
```diff
@@ -1,121 +1,91 @@
...
+    value: vestingsContracts,
+  } = useVesting_getVestingsOf(account);
+  const [possibleVestings, setPossibleVestings] = useState<Vesting[]>([]);
+
+  useEffect(() => {
+    async function getVestings() {
+      try {
+        const address: string[] = [],
+          type: string[] = [],
+          typeCreation: string[] = [];
```
ShynRou, line 24: Please don't use chained declarations.

```suggestion
        const address: string[] = [];
        const type: string[] = [];
        const typeCreation: string[] = [];
```

src/app/components/UserAssets/Vesting/useListOfUserVestings.ts
```diff
@@ -1,121 +1,91 @@
...
+        if (!loadingVestings) {
+          for (let i in vestingsContracts) {
```
ShynRou, line 26: Probably no issue here, but be careful with for-in loops they are notorious for causing issues. For-in loops iterate over all object fields not over indexes, with a standard Array its no problem with an extended one it would iterate over function names.

src/app/components/UserAssets/Vesting/useListOfUserVestings.ts
```diff
@@ -1,121 +1,91 @@
...
-        );
-
-        return filteredVestings
-          .map((item, index) => {
-            item.vestingContract = contracts[index];
-            return item;
-          })
-          .filter(item => item.vestingContract !== ethGenesisAddress);
+        return possibleVestings as FullVesting[];
       };
```
ShynRou, line 88: A) Completely remove the run function, as it's unnecessary and will clean up the code below.
B) Do not use **as**. Type casting this way should not be done, since it disables typescript as a safety tool. Typing should be done at the first layer where an object is generated defined, in this case probably useVesting_getVestingsOf.

ShynRou (changes requested):

src/app/components/UserAssets/Vesting/useListOfUserVestings.ts
pietro-maximoff: @ShynRou Thanks for reviewing, i added refactoring based on your comments

PR #1792, opened 2021-10-08
Feat/440 perpetual order card

ShynRou:
Partial implementation of https://taiga.sovryn.app/project/sovryn-light/task/440

Missing all data fetching and the order modal.
Changed the layout to the new design.


t-z-u:
> I don't like the fact that `Position Leverage` selector jumps when you switch between `Market` and `Limit` type of orders. I would recommend to have `Position Leverage` selector at the bottom (above the submit button) at all times. Discuss it with @t-z-u please :)

this works exactly as expected!

tiltom:
Adjust this as well (spacings and link styling).

Current state:
image

Design:
image

PR #1834, opened 2021-10-27
Feat/508 swap data attributes

JoeStorm99:
https://taiga.sovryn.app/project/sovryn-light/task/508


soulBit (changes requested):
See Taiga ticket for requested changes

PR #1849, opened 2021-11-01
Feat/perpetuals/graphql

BetsyBraddock:
4 hooks for getting historical data from the Sovryn subgraph:

1. useGetTraderEvents - this hook returns past event data for user events (eg Trade, TokensWithdrawn). I can easily add more events into the Event enum here if needed

2. useGetCandles - returns historical data for candlestick chart

3. useGetRecentTrades - return historical data for the right hand side recent trade chart. For events after page load, websocket connection to the node can be used

4. useGetBannerData - returns data needed for the banner above the chart, for each perpetualId. Right now this is just the mark price and 24h volume data, but I can add other event-based data as needed

I was testing these with console logs in 2 test components TestHistory and TestCandles, which I've left in for reference


package.json
```diff
@@ -35,6 +37,7 @@
     "@typescript-eslint/parser": "^4.29.3",
     "@visx/responsive": "1.1.0",
     "animate.css": "4.1.1",
+    "apollo-client": "2.6.10",
```
ShynRou, line 40: Is apollo-client needed, when @apollo/client is already used?

BetsyBraddock: Good point. There may also be a few other un-needed packages, I'll double check and uninstall

tiltom (approved):

package.json
BetsyBraddock: @ShynRou it was needed - I think it's used by @apollo/client. I removed it and the code wouldn't compile. Same with the graphql package

ShynRou (changes requested):

package.json
ShynRou: Pushed it. Will merge once cloudflare is done.

ShynRou (approved):

PR #1856, opened 2021-11-03
Adjust openPosition hook and order form

tiltom:


src/app/pages/PerpetualPage/hooks/usePerpetual_openTrade.ts
```diff
@@ -1,28 +1,57 @@
...
+        leverage,
+        marginBalance.fPositionBC,
+        amount,
+        perpetualParameters,
+      );
 
-      if (!isClosePositionTransaction) {
-        await deposit(marginAmount);
+      if (!isClosePosition) {
+        await deposit(String(marginCollateralAmount));
```
ShynRou, line 54: Is marginCollateralAmount a wei amount? Otherwise you need to use toWei as deposit expects a wei amount, doesn't it?

ShynRou (changes requested):

src/app/pages/PerpetualPage/hooks/usePerpetual_openTrade.ts
tiltom: `marginCollateralAmount` is already wei amount (but it's a number and not a string).

ShynRou (approved):

PR #1867, opened 2021-11-10
Origins Zero FE

remedcu:


src/app/pages/OriginsLaunchpad/pages/SalesDay/index.tsx
```diff
@@ -8,42 +16,35 @@ import { useIsConnected } from 'app/hooks/useAccount';
...
-        return <EngageWalletStep saleName={saleName} />;
+  useEffect(() => {
+    const { step } = saleStorage.getData();
+    if (connected) {
+      history.push(`${match.url}/${step}`);
+    } else {
+      history.push(`${match.url}/engage-wallet`);
     }
-  };
+    // eslint-disable-next-line
```
tiltom, line 46: Why do you have eslint ignore here?

src/app/pages/OriginsLaunchpad/pages/SalesDay/index.tsx
```diff
@@ -8,42 +16,35 @@ import { useIsConnected } from 'app/hooks/useAccount';
...
 import { useGetSaleInformation } from '../../hooks/useGetSaleInformation';
+import saleStorage from './storage';
 import styles from './index.module.scss';
 
 interface ISalesDayProps {
   saleName: string;
 }
 
 export const SalesDay: React.FC<ISalesDayProps> = ({ saleName }) => {
+  const match = useRouteMatch();
```
tiltom, line 27:
```suggestion
  const { url } = useRouteMatch();
```

src/app/pages/OriginsLaunchpad/index.tsx
```diff
@@ -1,13 +1,16 @@
...
+import { Switch, Route, Redirect, useRouteMatch } from 'react-router-dom';
 import { translations } from 'locales/i18n';
 import { Header } from 'app/components/Header';
 import { Footer } from 'app/components/Footer';
 import { SalesDay } from './pages/SalesDay';
+import { OriginsClaimPage } from 'app/pages/OriginsClaimPage/Loadable';
 
 export const OriginsLaunchpad: React.FC = () => {
   const { t } = useTranslation();
+  const match = useRouteMatch();
```
tiltom, line 13:
```suggestion
  const { url, path } = useRouteMatch();
```

src/app/pages/OriginsLaunchpad/pages/SalesDay/storage.ts
```diff
@@ -0,0 +1,22 @@
...
+const defaultData: ISalesData = {
+  step: 1,
+};
+
+const saleStorage = {
+  saveData: (data: ISalesData): void => {
+    local.setItem(originsSaleStorageKey, JSON.stringify(data));
+  },
+  getData: (): ISalesData => {
+    const strData = local.getItem(originsSaleStorageKey);
```
tiltom, line 17:
```suggestion
    const data = local.getItem(originsSaleStorageKey);
```

src/app/pages/OriginsLaunchpad/pages/SalesDay/storage.ts
```diff
@@ -0,0 +1,22 @@
...
+    return strData ? (JSON.parse(strData) as ISalesData) : defaultData;
```
tiltom, line 18:
```suggestion
    return strData ? (JSON.parse(strData)) : defaultData;
```

src/app/pages/OriginsLaunchpad/pages/SalesDay/storage.ts
```diff
@@ -0,0 +1,22 @@
+import { originsSaleStorageKey } from 'utils/classifiers';
```
tiltom, line 1: I am not sure if the whole idea of saving the active step to local storage is not overkill - we basically just want our users to be able to refresh without worries and that's done by routes. 

What do you think @soulBit ?

tiltom (changes requested):

src/app/pages/OriginsLaunchpad/pages/SalesDay/index.tsx
soulBit: Please make the requested changes to props pulled from `useRouteMatch()` further down this file, then remove this line and add `url` and `history` to dependency array to satisfy compiler.

src/app/pages/OriginsLaunchpad/pages/SalesDay/storage.ts
soulBit: I agree with tiltom on this comment, unless there was a specific reason it was added?

src/utils/classifiers.ts
```diff
@@ -48,6 +48,8 @@ export const sovAnalyticsCookie = { name: 'SovAnalytics', value: 'optout' };
 
 export const chartStorageKey = 'sovryn.charts';
 
+export const originsSaleStorageKey = 'sovryn.origins.sales';
```
soulBit, line 51: Remove this if local storage functionality is removed based on other suggested changes

soulBit (changes requested):
Please make requested changes

src/app/pages/OriginsLaunchpad/pages/SalesDay/index.tsx
business-helper: okay I got it.

src/app/pages/OriginsLaunchpad/pages/SalesDay/storage.ts
business-helper: when users refresh in the 3rd step, the wallet needs to be connected again, and then the routing system will lead the user to the previous page programmatically.
So I think it's still needed.

tiltom: Ok, if you need it there, it can stay the way it is.

soulBit (approved):
Zero token contract address needs to be added to backend by @BetsyBraddock to resolve price feed errors in console. This should also be done for mainnet once contracts have been deployed there.

In addition, we should remove the `useDollarValueZero()` hook call from `src/app/components/UserAssets/index.tsx` once bootstrap event is over (in the same way we have done for Mynt once bonding curve contract is live).

PR #1870, opened 2021-11-10
feat: add deposit button to header, xusd asset icons to portfolio, re…

soulBit:
…move action button borders on portfolio

https://taiga.sovryn.app/project/sovryn-light/us/627?milestone=40

### DIFF
--- a/src/app/components/UserAssets/index.module.scss
+++ b/src/app/components/UserAssets/index.module.scss
@@ -1,0 +1,5 @@
+.actionLink {
+  @apply tw-text-primary tw-tracking-normal hover:tw-text-primary hover:tw-underline tw-mr-1 xl:tw-mr-4 tw-p-0 tw-font-normal;
+
+  font-family: Montserrat;
+}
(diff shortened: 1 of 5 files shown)


JoeStorm99 (dismissed):

src/app/components/UserAssets/index.module.scss
```diff
@@ -0,0 +1,5 @@
+.actionLink {
+  @apply tw-text-primary tw-tracking-normal hover:tw-text-primary hover:tw-underline tw-mr-1 xl:tw-mr-4 tw-p-0 tw-font-normal;
+
+  font-family: Montserrat;
```
tiltom, line 4: There is no need to specify `font-family` like this because we already have `Montserat` defined in `tailwind.config.ts` so you could just add `tw-font-body` to `@apply`.

Since you have no custom property here, you could delete this selector and the whole file (but I can see it's used on 4 places so I'll leave it up to you).

tiltom (changes requested):

tiltom:
I think that spacing between `Deposit` and `Buy` in `rBTC` is not proportional to spacing used in other assets.

image

tiltom (approved):

PR #1908, opened 2021-11-26
fix: reward SOV shown for non-LM Lend pools

soulBit:
https://taiga.sovryn.app/project/sovryn-light/issue/665

Bug fix for cases where user has accrued reward SOV for other liquidity mining enabled pools, and rewards appear for non-LM enabled pools such as RBTC. Issue is caused by failed call to `liquidityMiningProxy.getPoolId()`, which was using default value of 0 (thereby causing reward SOV value to be retrieved for another pool).


tiltom (dismissed):

Alitasovryn (dismissed):

soulBit:
Merged via #1927

PR #1914, opened 2021-11-30
Use a context for perpetual queries

tiltom:


src/app/pages/PerpetualPage/contexts/PerpetualQueriesContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(
+    () =>
+      setValue({
+        ammState,
+        perpetualParameters,
+        traderState,
+        liquidityPoolState,
+      }),
+    [ammState, liquidityPoolState, perpetualParameters, traderState],
+  );
```
ShynRou, line 63: The state and effect combination is not needed here, you can just use useMemo.
with effect the provider needs to be called twice by react.

ShynRou (changes requested):

ShynRou (approved):

PR #1916, opened 2021-12-02
Update perpetual manager

tiltom:


src/app/pages/PerpetualPage/hooks/usePerpetual_AmmDepthChart.ts
```diff
@@ -29,20 +24,27 @@ export type AmmDepthChartData = {
...
 
     let shorts: AmmDepthChartDataEntry[] = [];
     let longs: AmmDepthChartDataEntry[] = [];
 
     if (entries && entries.length >= 3) {
       const length = entries[0].length;
       const midIndex = Math.floor(length / 2);
+
+      const averagePriceIndex = entries[1].indexOf(0);
+      averagePrice = entries[0][averagePriceIndex];
```
ShynRou, line 46: your averagePriceIndex should be the same as midIndex.
either use midIndex or your indexing approach. not both.
```suggestion
      averagePrice = entries[0][midIndex];
```

src/app/pages/PerpetualPage/hooks/usePerpetual_openTrade.ts
```diff
@@ -25,14 +25,25 @@ const MASK_CLOSE_ONLY = 0x80000000;
...
     perpetualParameters,
     ammState,
   ]);
 
+  let averagePrice = undefined;
+
+  if (entries && entries.length >= 3) {
+    const averagePriceIndex = entries[1].indexOf(0);
+    averagePrice = entries[0][averagePriceIndex];
+  }
```
ShynRou, line 45: I really dislike this solution. We will also need to implement it in the trade form and close modal to show the correct limit values. Either we add a utils function or just add the extracted averagePrice to the context

ShynRou (changes requested):

src/app/pages/PerpetualPage/hooks/usePerpetual_openTrade.ts
tiltom: I decided to add it to the context and I also used it in the trade form and slippage modals.

ShynRou (approved):

PR #1920, opened 2021-12-06
Update perpetuals UI on new trades

tiltom:


src/app/pages/PerpetualPage/hooks/usePerpetual_getLatestTradeId.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const usePerpetual_getLatestTradeId = () => {
+  const [latestTradeId, setLatestTradeId] = useState<Nullable<string>>();
+  const { trades } = useContext(RecentTradesContext);
+
+  useDebouncedEffect(
+    () => setLatestTradeId(trades[0]?.id),
+    [trades],
+    BSC_AVERAGE_BLOCK_TIME,
+  );
```
ShynRou, line 16: We don't need debounced but throttled.
Debounced delays as long as something comes in.
Throttled instantly calles, and then will only update again after a specific delay.

ShynRou (approved):

PR #1924, opened 2021-12-07
Order History table data

tiltom:


src/app/pages/PerpetualPage/hooks/graphql/useGetTraderEvents.ts
```diff
@@ -28,19 +38,34 @@ function generateQuery(
...
+      isOrdered || hasPagination
+        ? `(${orderFilterString} ${paginationFilterString})`
         : ''
     } { ${eventDetails.fields.toString()} }`;
   });
 
   return gql`
   { trader(id: "${user}")
   {
+    ${getTotalCountFieldName(events[0])}
```
ShynRou, line 68: This should also map all event types in the array, and filter out unsupported ones.

src/app/pages/PerpetualPage/hooks/graphql/useGetTraderEvents.ts
```diff
@@ -57,6 +82,15 @@ export enum Event {
...
 }
 
+const getTotalCountFieldName = (event: Event) => {
+  switch (event) {
+    case Event.TRADE:
+      return 'tradesTotalCount';
+    default:
+      return 'tradesTotalCount';
+  }
+};
```
ShynRou, line 92: No function needed here you can just add an object map that uses the Event enum as key. There should be no default value as the count is only supported for the trades and positions right now.

ShynRou (changes requested):

ShynRou (approved):

PR #1925, opened 2021-12-07
fix: bump gas limit for staking

soulBit:
Increase gas limit for staking SOV - this fixes gas limit issues for HW wallet users, please use a HW wallet to test this.

AC:
* User is able to stake SOV without receiving a failed tx with "out of gas" error when using HW wallet

Negative path:
* User cannot stake SOV and receives a failed tx with "out of gas" error when using HW wallet


soulBit:
Merged via #1927

PR #1971, opened 2021-12-22
[Perpetual] call contracts with bridgeNetwork

ShynRou:


src/app/hooks/useBridgeNetworkSendTx.ts
```diff
@@ -0,0 +1,160 @@
...
+  reset: () => void;
+}
+
+export interface SendTxResponseInterface extends ResetTxResponseInterface {
+  send: (
+    args: any[],
+    config?: TransactionConfig,
+    options?: TransactionOptions,
+  ) => void;
+}
```
grinry, line 48: I think these could be deleted and reused from useContractSendTx hook

grinry (changes requested):

src/app/hooks/useBridgeNetworkSendTx.ts
ShynRou: Removed them and imported from useContractSendTx

creed-victor (approved):

PR #1982, opened 2022-01-04
Feat/685 Perpetual max trade sizes and balance validation

ShynRou:
Resolves https://taiga.sovryn.app/project/sovryn-light/task/685


src/app/pages/PerpetualPage/utils/contractUtils.tsx
```diff
@@ -257,5 +247,32 @@ export const validatePositionChange = (
...
+      ammState,
+      traderState,
+      slippage,
+      useMetaTransactions,
+    );
+
+    if (requiredMargin > traderState.availableCashCC + availableBalance) {
+      result.valid = false;
+      result.isWarning = true;
+      result.errors.push(new Error('Order cost exceeds total balance!'));
```
tiltom, line 266: We need to use localized strings.

tiltom (changes requested):

src/app/pages/PerpetualPage/utils/contractUtils.tsx
ShynRou: The error message here is just for logging not for displaying purposes, that's what the result.errorMessages are for.

PR #1983, opened 2022-01-11
Feat/694 Perpetual second pair preperation

ShynRou:
Resolves https://taiga.sovryn.app/project/sovryn-light/task/694


src/app/pages/PerpetualPage/contexts/PerpetualQueriesContext.tsx
```diff
@@ -164,6 +177,26 @@ export const PerpetualQueriesContextProvider: React.FC<PerpetualQueriesContextPr
...
+            .call(
+              Chain.BSC,
+              getContract('PERPETUALS_token').address,
+              marginTokenAbi,
+              'balanceOf',
+              [account],
+            )
+            .then(result => result && setAvailableBalance(String(result)))
+            .catch(console.error),
+        500,
```
tiltom, line 194: Why do you have `500 ms` here? We use `1000 ms` (`THROTTLE_DELAY`) everywhere else.

src/app/pages/PerpetualPage/contexts/PerpetualQueriesContext.tsx
```diff
@@ -164,6 +177,26 @@ export const PerpetualQueriesContextProvider: React.FC<PerpetualQueriesContextPr
...
+        { leading: false, trailing: true, maxWait: 500 },
```
tiltom, line 195: `leading: false` and `traling: true` are the default values (according to the [documentation](https://lodash.com/docs/4.17.15#debounce)) so there is no need to specify them.

src/app/pages/PerpetualPage/contexts/PerpetualQueriesContext.tsx
```diff
@@ -128,29 +143,27 @@ export const PerpetualQueriesContextProvider: React.FC<PerpetualQueriesContextPr
...
-    debounce(() => refetchTraderState(), THROTTLE_DELAY, {
-      leading: true,
-      trailing: true,
-      maxWait: THROTTLE_DELAY,
-    }),
+  const refetchUser = useMemo(
+    () =>
+      debounce(() => refetchTraderState(), THROTTLE_DELAY, {
+        leading: true,
+        trailing: true,
```
tiltom, line 164: `true` is the default value of `trailing` so there is no need to specify it.

tiltom (changes requested):

src/app/pages/PerpetualPage/contexts/PerpetualQueriesContext.tsx
ShynRou: Idk sounded like a reasonable value :D

PR #1988, opened 2022-01-13
fix(bridge): reset USDT allowance to 0

creed-victor:
Resets USDT allowance to 0 before requesting approval for more (ethereum side)

QA:
- Open portfolio
- Select XUSD -> Deposit
- Choose USDT as deposit token
- Enter amount you want to deposit (do not use entire balance)
- Confirm transaction for approval tx
- Cancel second transaction (the deposit one)
- Repeat it for second time but do not cancel transaction anymore and with higher amount than first time: you should be asked to confirm 3 transactions (2 approval, 1 deposit)
- Repeat it for third time: you should be asked to confirm 2 transactions (1 approval, 1 deposit)

Try same thing for any other token than USDT too - you should be asked to confirm 2 transactions and not 3 for other ones.

### DIFF
--- a/src/app/pages/BridgeDepositPage/saga.ts
+++ b/src/app/pages/BridgeDepositPage/saga.ts
@@ -30,6 +30,7 @@ import { bignumber } from 'mathjs';
 import { ethers } from 'ethers';
 import { TxType } from 'store/global/transactions-store/types';
 import { gasLimit } from 'utils/classifiers';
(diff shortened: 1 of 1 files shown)


soulBit (approved):
Worked well for USDT and USDC 👍🏻

Alitasovryn (approved):
Tried with USDT. Working well

PR #1997, opened 2022-01-17
Fix slider reset with custom value

tiltom:


ShynRou:
I don't think this works for us, since the labels are crammed already... I will implement a simpler approach.
This would be a good solution, if we had the space under the slider to add the custom entry.

ShynRou (approved):

PR #1999, opened 2022-01-17
Show approximate and the worst case price

tiltom:


ShynRou (changes requested):
Code looks good. I will adjust the spacing a bit and flip the TradeSummary close value label pairs.

PR #2002, opened 2022-01-19
Feat/703 Perpetuals Graph Fallback

ShynRou:
Resolves https://taiga.sovryn.app/project/sovryn-light/task/703

Will need to be used by the Competition page as well.

### DIFF
--- a/src/app/pages/PerpetualPage/contexts/GraphQLProvider.tsx
+++ b/src/app/pages/PerpetualPage/contexts/GraphQLProvider.tsx
@@ -1,0 +1,213 @@
+import React, { useState, useEffect } from 'react';
+import { ApolloClient, InMemoryCache, ApolloProvider } from '@apollo/client';
+
+type GraphQLEndpoint = { graph: string; index?: string; isFallback?: boolean };
+
+const metaQuery = `{
+  _meta{
+    block {
+      hash
+    }
+    deployment
+    hasIndexingErrors
+  }
+}`;
+
+const statusQuery = deployment => `{
+  indexingStatuses(subgraphs: [${JSON.stringify(deployment)}]) {
+    subgraph
+    synced
+    health
+    entityCount
+    fatalError {
+      handler
+      message
+      deterministic
+    }
+    chains {
+      chainHeadBlock {
+        number
+      }
+      earliestBlock {
+        number
+      }
+      latestBlock {
+        number
+      }
+    }
+  }
+}`;
+
+const createClient = (endpoint: GraphQLEndpoint) =>
+  new ApolloClient({
+    uri: endpoint.graph,
+    cache: new InMemoryCache({
+      resultCaching: false,
+    }),
+  });
+
+type EvaluationResult = {
+  score: number;
+  isUp?: boolean;
+  isSynced?: boolean;
+  isHealthy?: boolean;
+};
+
+const evaluateEndpoint = async (endpoint: GraphQLEndpoint) => {
+  const tests = 3;
+  let score = 0;
+
+  let result: EvaluationResult = {
+    score: 0,
+    isUp: false,
+    isSynced: undefined,
+    isHealthy: undefined,
+  };
+
+  try {
+    const metaResponse = await (
+      await fetch(endpoint.graph, {
+        method: 'POST',
+        body: JSON.stringify({ query: metaQuery }),
+      })
+    ).json();
+
+    if (!metaResponse?.data?._meta?.deployment) {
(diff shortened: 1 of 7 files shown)


creed-victor (approved):

src/app/pages/PerpetualPage/contexts/GraphQLProvider.tsx
```diff
@@ -0,0 +1,210 @@
...
+  isHealthy?: boolean;
+};
+
+const evaluateEndpoint = async (endpoint: GraphQLEndpoint) => {
+  const tests = 3;
+  let score = 0;
+
+  let result: EvaluationResult = {
+    score: 0,
+    isUp: false,
```
tiltom, line 62: Set the default value to `undefined` or adjust the type declaration.

src/app/pages/PerpetualPage/contexts/GraphQLProvider.tsx
```diff
@@ -0,0 +1,210 @@
...
+      console.error('Failed to fetch graph deployment id!', metaResponse);
+      return result;
+    }
+
+    score++;
+    result.isUp = true;
+
+    if (!endpoint.index) {
+      // endpoint can not be evaluated further
+      result.score = score / tests;
```
tiltom, line 85: I don't think you need to divide `score` by number of `tests` (you would just need to adjust lines `153` and `179`).

This is just an opinion so I'll leave it up to you if you want to change it but you would get rid of one variable and 2 divisions.

src/app/pages/PerpetualPage/contexts/GraphQLProvider.tsx
```diff
@@ -0,0 +1,210 @@
...
+    index: 'https://api.thegraph.com/index-node/graphql',
+  },
+  {
+    graph:
+      'https://graphql.sovryn.app/subgraphs/name/DistributedCollective/Sovryn-perpetual-swaps-subgraph',
+    isFallback: true,
+  },
+];
+
+const RETRY_DELAY = 2 * 60 * 1000;
```
tiltom, line 131:
```suggestion
const RETRY_DELAY = 2 * 60 * 1000; // 2 minutes
```

src/app/pages/PerpetualPage/contexts/GraphQLProvider.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+type GraphQLProvicerProps = {
```
tiltom, line 133:
```suggestion
type GraphQLProviderProps = {
```

src/app/pages/PerpetualPage/hooks/usePerpetual_ContractDetails.ts
```diff
@@ -49,7 +50,13 @@ export const usePerpetual_ContractDetails = (pairType: PerpetualPairType) => {
     };
 
     get24hVolume().catch(console.error);
-  }, [pair.id]);
+    const intervalId = setInterval(
+      () => get24hVolume().catch(console.error),
+      5 * 60 * 1000,
```
tiltom, line 55: Maybe you can create a global constant for 1 minute and use it here and in `GraphQLProvider` as well (you wouldn't need comments then).

tiltom (changes requested):

src/app/pages/PerpetualPage/contexts/GraphQLProvider.tsx
ShynRou: I added it, because the original thought was to have percentage scrores, so you could calculate in the exact amount of blocks behind head and compare multiple nodes, but it might not really be needed.

src/app/pages/PerpetualPage/hooks/usePerpetual_ContractDetails.ts
ShynRou: We could move this to a constant as well, but I don't think a constant for one minute in ms is really needed.

src/app/pages/PerpetualPage/contexts/GraphQLProvider.tsx
```diff
@@ -128,13 +123,14 @@ const config: GraphQLEndpoint[] = [
   },
 ];
 
-const RETRY_DELAY = 2 * 60 * 1000;
+const RETRY_DELAY = 2 * 60 * 1000; // 2 seconds
```
ShynRou, line 126: This should be 2 minutes
```suggestion
const RETRY_DELAY = 2 * 60 * 1000; // 2 minutes
```

PR #2023, opened 2022-02-02
Remove leverage from SOV margin trade

tiltom:
Resolves https://sovryn.monday.com/boards/2218344956/pulses/2231505190

### DIFF
--- a/src/app/pages/MarginTradePage/components/TradeForm/index.tsx
+++ b/src/app/pages/MarginTradePage/components/TradeForm/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { useTranslation, Trans } from 'react-i18next';
 import { translations } from '../../../../../locales/i18n';
 import { Select } from 'app/components/Form/Select';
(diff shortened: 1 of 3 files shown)


soulBit (approved):
Refactored SOV leverage const name and added small explanatory comment - changes look good 👍🏻

creed-victor (dismissed):

creed-victor:
Changes ok, but we could improve it with trading pair dictionary and add available leverage as a config there instead (in case there will be more pairs with different leverage options in the future).

tiltom:
> Changes ok, but we could improve it with trading pair dictionary and add available leverage as a config there instead (in case there will be more pairs with different leverage options in the future).

I thought this is a one-off thing that's why I made it this way but I can add it to trading pair dictionary so it's more universal.

creed-victor (approved):

creed-victor:
> I thought this is a one-off thing that's why I made it this way but I can add it to trading pair dictionary so it's more universal.

Should be one-off thing, but I remember changing leverage min/max a year ago :D Thanks

JoeStorm99 (approved):
tested and all looks fine but need to complete testing on staging due to an intermittent error on "price disagreement".

soulBit (approved):

PR #2047, opened 2022-02-14
[Spinner] - update spinner

pietro-maximoff:
https://sovryn.monday.com/boards/2218344956/pulses/2255718606


src/app/components/Spinner/index.module.scss
```diff
@@ -0,0 +1,15 @@
+.spinner {
+    @apply tw-flex tw-items-center tw-justify-center tw-text-gray-7;
+    svg {
+        min-width: 10px;
+        min-height: 10px;
+        max-width: 100px;
+        max-height: 100px;
```
tiltom, line 7:
```suggestion
        min-width: 0.625rem;
        min-height: 0.625rem;
        max-width: 6.25rem;
        max-height: 6.25rem;
```

src/app/components/Spinner/index.module.scss
```diff
@@ -0,0 +1,15 @@
...
+        path {
+            &:first-child {
+                @apply tw-opacity-25;
+                stroke: rgb(92, 112, 128);
```
tiltom, line 11: Why do we use a color that is not in our color paletter (in `tailwind.config.js`)? Try to ask tzu if he's ok with that.

src/app/components/Spinner/index.tsx
```diff
@@ -1,28 +1,37 @@
 import React from 'react';
-import { Spinner as BPSpinner } from '@blueprintjs/core/lib/esm/components/spinner/spinner';
 import cn from 'classnames';
```
tiltom, line 2:
```suggestion
import classNames from 'classnames';
```

src/app/components/Spinner/index.tsx
```diff
@@ -1,28 +1,37 @@
...
-interface Props {
+interface ISpinnerProps {
   className?: string;
-  innerClassName?: string;
+  size?: SpinnerSizeType;
 }
 
-export const Spinner: React.FC<Props> = ({ className, innerClassName }) => {
+export const Spinner: React.FC<ISpinnerProps> = ({ className, size }) => {
   return (
```
tiltom, line 12:
```suggestion
export const Spinner: React.FC<ISpinnerProps> = ({ className, size = SpinnerSizeType.STANDARD }) => (
```

src/app/components/Spinner/index.tsx
```diff
@@ -1,28 +1,37 @@
...
-      <span
-        className={cn(
-          'tw-w-min tw-btn-loader__spinner tw-flex tw-flex-row tw-items-center active',
-          innerClassName,
-        )}
+    <div className={cn(styles.spinner, className)}>
+      <svg
+        className="tw-animate-spin-fast"
+        width={size || SpinnerSizeType.STANDARD}
+        height={size || SpinnerSizeType.STANDARD}
```
tiltom, line 17:
```suggestion
        width={size}
        height={size}
```

tiltom (changes requested):

src/app/components/Spinner/index.tsx
pietro-maximoff: +

src/app/components/Spinner/index.module.scss
pietro-maximoff: this color was taken from the blueprintjs/spinner

src/app/components/Spinner/index.module.scss
pietro-maximoff: changed, we will use "rem" everywhere instead of "px"?

src/app/components/Spinner/index.tsx
pietro-maximoff: fixed

src/app/components/Spinner/index.tsx
pietro-maximoff: set to the standard size in props by default

src/app/components/Spinner/index.module.scss
tiltom: Ok but still, verify it with him pls. I don't dispute the fact it was in blueprintjs that way but since we are building our custom component, I believe we should customize them so they fit our design system.

src/app/components/Spinner/index.module.scss
tiltom: Almost everywhere :D Or better put - everywhere where it makes sense (and it does make sense for almost everything but you don't want for instance border width to be wider on bigger screens so it's completely ok to use `px` there).

tiltom (approved):

src/app/components/Spinner/index.module.scss
ShynRou: Just change it to the closest matching gray-x

pietro-maximoff: yup, I did even better - I used the same color but with 90% transparency, it gives the same effect and looks much more harmonious if you make it colored

tiltom: Looking good @pietro-maximoff :)

tiltom (approved):

ShynRou (approved):

creed-victor (approved):

JoeStorm99 (approved):

soulBit (approved):

PR #2068, opened 2022-02-22
Feat/margin limit order

rick23p:


src/app/pages/MarginTradePage/components/TradeForm/index.tsx
```diff
@@ -309,6 +376,14 @@ export const TradeForm: React.FC<ITradeFormProps> = ({ pairType }) => {
           )}
         </div>
 
+        {openLimitSetting && (
+          <LimitOrderSetting
+            onClose={() => setOpenLimitSetting(false)}
+            duration={duration}
+            handleDuration={value => setDuration(value)}
```
tiltom, line 383:
```suggestion
            handleDuration={setDuration}
```

tiltom (changes requested):

tiltom:
Review Order dialog needs adjustments - there are too many decimal places everywhere (the most notable is `Minimum Entry Price`) + margin button and the content should have the same width (I would say that the content width needs to be adjusted because it's unnecessary wide).

image

tiltom:
Cancel Limit Order dialog needs adjustments as well - shrink content width to the button width.

image

PR #2073, opened 2022-02-25
Checkbox atomic component

tiltom:
Resolves https://sovryn.monday.com/boards/2218344956/pulses/2255710988

### DIFF
--- a/src/app/components/Checkbox/index.stories.tsx
+++ b/src/app/components/Checkbox/index.stories.tsx
@@ -1,0 +1,18 @@
+import React, { ComponentProps } from 'react';
+import { Story } from '@storybook/react';
+
+import { Checkbox } from './index';
+
+export default {
+  title: 'Atoms/Checkbox',
+  component: Checkbox,
+};
+
+const Template: Story<ComponentProps<typeof Checkbox>> = args => (
+  <Checkbox {...args} />
+);
+
+export const Basic = Template.bind({});
+Basic.args = {
+  label: 'This is a simple checkbox',
+};
--- a/src/app/components/Checkbox/index.stories.tsx
+++ b/src/app/components/Checkbox/index.stories.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentProps } from 'react';
+import React, { ComponentProps, useCallback, useState } from 'react';
 import { Story } from '@storybook/react';
 
 import { Checkbox } from './index';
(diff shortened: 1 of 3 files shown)


src/app/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+import React, { useCallback, useMemo } from 'react';
+import styles from './index.module.scss';
+
+export enum CheckboxAlignment {
+  LEFT = 'left',
+  RIGHT = 'right',
+}
+
+type CheckboxProps = {
+  label: string | React.ReactNode;
```
ShynRou, line 11:
```suggestion
  label: React.ReactNode;
```
string is already included in ReactNode

src/app/components/Checkbox/index.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Checkbox } from './index';
+
+export default {
+  title: 'Atoms/Checkbox',
+  component: Checkbox,
+};
+
+const Template: Story<ComponentProps<typeof Checkbox>> = args => (
+  <Checkbox {...args} />
+);
```
ShynRou, line 13: Having a second interactive template as well might be good. So a template that handles the checked state.

src/app/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  onChange,
+  disabled = false,
+  inline = false,
+  alignment = CheckboxAlignment.LEFT,
+  className,
+}) => {
+  const onClick = useCallback(
+    e => {
+      e.preventDefault();
+      !disabled && onChange();
```
ShynRou, line 32: Please just use an `if`. Also please pass the new value to onChange.

ShynRou (changes requested):

src/app/components/Checkbox/index.tsx
tiltom: There is no need for passing values, `checked` is controlled by a parent component (but good catch with `if`).

src/app/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+        className,
+      )}
+    >
+      <label
+        className={classNames(styles.label, {
+          [styles.labelDisabled]: disabled,
+        })}
+        onClick={onClick}
+      >
+        <input type="checkbox" className="tw-hidden" />
```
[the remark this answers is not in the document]
tiltom: Yes, you're right, `onChange` needs to be used on the input instead of `onClick` on the label.

creed-victor (dismissed):

GiladSh05:
1 small design feedback I have is that personally I would prefer the "V" inside the checkbox when selected to be black

tiltom:
> 1 small design feedback I have is that personally I would prefer the "V" inside the checkbox when selected to be black

I made it black, pls review it again if it looks good (after new deployment is done).

ShynRou (approved):

soulBit (approved):

PR #2075, opened 2022-02-25
Feat/mixpanel test

creed-victor:
https://sovryn.monday.com/boards/2218344956/pulses/2338467593


creed-victor:
feature was dropped

PR #2079, opened 2022-03-02
Spinner enforcement

creed-victor:
https://sovryn.monday.com/boards/2218344956/pulses/2324277884

AC: new spinner is working in all old places (example withdraw button in vested assets list)


creed-victor:
> The PR itself looks good but we still have plenty of style overrides for the bp3 spinner, please delete them. Also, I believe that the `trade-btn.css` file can be completely deleted since we no longer have the `TradeButton` component.

removed classes but left teade-btn.scss file - classes are still used for some buttons in margin trading

tiltom (approved):

src/app/components/SalesButton/index.tsx
```diff
@@ -64,7 +64,7 @@ export default function SalesButton({
   return (
     <StyledButton onClick={() => onClick()} disabled={disabled}>
       {text}
-      {loading && <Spinner size={18} className="tw-ml-2" />}
+      {loading && <Spinner size={20} className="tw-ml-2" />}
```
ShynRou, line 67: Could you please use the enum instead of the enum.

edit: looking at the enum could you also adjust the name to SpinnerSize and the sizes to SM, MD, LG, to fit the usual pattern

ShynRou (changes requested):

soulBit (approved):

PR #2084, opened 2022-03-03
Design changes

tiltom:
Resolves https://sovryn.monday.com/boards/2218344956/pulses/2353004014


ShynRou (changes requested):
Could you change it to resemble this.
image

Also the DataCards don't fill the whole width anymore with this changes.

PR #2088, opened 2022-03-08
Feat/2236012071 tabs atomic component

rick23p:
https://sovryn.monday.com/boards/2218344956/pulses/2236012071


src/app/components/Tabs/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  initial,
+  onChange,
+  className,
+  contentClassName,
+}) => {
+  const [activeTab, setActiveTab] = useState(initial);
+
+  const selectTab = (item: ITabItem) => {
+    if (item.disabled) return;
+    setActiveTab(item.id);
```
tiltom, line 30:
```suggestion
    if (item.disabled) {
      return;
    }
    setActiveTab(item.id);
```

src/app/components/Tabs/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const Tabs: React.FC<TabsProps> = ({
+  items,
+  initial,
+  onChange,
+  className,
+  contentClassName,
+}) => {
+  const [activeTab, setActiveTab] = useState(initial);
+
+  const selectTab = (item: ITabItem) => {
```
tiltom, line 28: Use `useCallback`.

tiltom (changes requested):

tiltom (dismissed):

pietro-maximoff:
just a quick suggestion here, maybe make sense to add an option that will allow you to combine all tabs into one select element based on media breakpoints, its more useful for mobile devices.
for instance if we have 3 or more tabs then we can hide them on mobile and show one select element, that's will be more conveniently.
Screenshot at Mar 10 23-22-01

JoeStorm99 (approved):

ShynRou:
@pietro-maximoff That's a good point, but I think that should be something we address at a later stage.

src/app/components/Tab/index.tsx
```diff
@@ -1,41 +1,33 @@
...
+  disabled,
+  className,
+}) => (
+  <span
+    className={classNames(
+      'tw-text-sov-white tw-px-2.5 tw-py-2 tw-text-lg tw-font-thin tw-whitespace-nowrap',
+      {
+        'tw-text-white tw-font-normal': active,
+        'tw-cursor-pointer hover:tw-text-gray-9': !disabled,
+        'tw-opacity-50': disabled,
```
ShynRou, line 25:
```suggestion
        'tw-opacity-50 tw-cursor-not-allowed': disabled,
```

src/app/components/Tabs/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <div className={className}>
+      <div className="tw-flex tw-items-center tw-flex-nowrap tw-overflow-auto">
+        {items.map(item => (
+          <Tab
+            key={item.id}
+            active={item.id === activeTab}
+            disabled={item.disabled}
+            onClick={() => selectTab(item)}
+            text={item.label}
+            className="tw-mr-2"
```
ShynRou, line 54:
```suggestion
            className="tw-mx-2"
```

src/app/components/Tab/index.tsx
```diff
@@ -1,41 +1,33 @@
...
-    `}
-`;
+export const Tab: React.FC<TabProps> = ({
+  text,
+  active,
+  onClick,
+  disabled,
+  className,
+}) => (
+  <span
```
ShynRou, line 19: Can you please make this into a button, so it is focus-able and shows up on screen readers.

ShynRou (changes requested):

soulBit (dismissed):
Reviewed code and resolved previous comments

PR #2095, opened 2022-03-14
[Perpetuals] disable perpetuals on mainnet

ShynRou:
Resolves https://sovryn.monday.com/boards/2218344956/pulses/2399662369

Removed Perpetuals from router and navigation.

@creed-victor I did not make any changes to FastBTC so navigating to /fast-btc/withdraw/bsc still allows for btc to bsc transfers and links to perps. Though there is no link to this path in the dapp. Do you think that could cause problems or can we leave it in in the meantime?


creed-victor (approved):

creed-victor:
Backend was upgraded to support btc deposits to bsc so even if user would find it deposit he would not lose his money, so i think it's fine to leave it be.
Ofc he may not be able to withdraw back - not sure if mainnet smart-contracts are added to FE contract list

soulBit (approved):

PR #2098, opened 2022-03-15
universal header component (part 1)

creed-victor:
A page container component which wraps router and has power to enable/disable header and footer, also swap default header or footer components from within page.
Leaving page where options were customised will reset settings back to defaults.

Usage (if needed to change header, footer or pass props to header from within page):

```tsx
const SomeCustomPage = () => {
   const page = usePageActions();
   useLayoutEffect(() => {
      // all options are optional
      page.setOptions({
         header: HeaderType.CROSS_CHAIN,
         footer: FooterTypes.NONE,
         headerProps: { address: receiver },
      });
   }, [page]);
  return (
     <span>Hello</span>
  );
};
```

part 2 will be PR to perpetuals with header for Labs.


JoeStorm99:
The following pages have a different header, however, I am aware of FastBTC should have a different header this is clearly stated on the PR but the Governance and 


Screenshot 2022-03-15 at 16 34 58
the Forum page and the Help has a different header: 


Screenshot 2022-03-15 at 16 32 06

Screenshot 2022-03-15 at 16 25 59

ShynRou (changes requested):
I'm sorry, but I don't think your solution to use redux for it is any good.
1. Functions/Components should never be passed into the redux store.
2. Each header change will need to be called by the component that requests the change, this means that a) all pages need to call the set function and b) the async nature of the pages will result in the header change only happening after the new page has been loaded.

I would suggest using the react-router for it, which eliminates all issues with handling the page changes. If you want to have the configuration in the page files instead of one centralized place you could add a register function, that is called statically outside of the page components.

ShynRou:
@JoeStorm99 None of the pages listed by you are part of this repository and with that out of the scope of this task.

src/app/containers/PageContainer/context/actions.ts
```diff
@@ -0,0 +1,23 @@
...
+export const useActions = (dispatch: Dispatch<PageContextAction>) => {
+  const memoized = useMemo(
+    () => ({
+      setOptions: (options: Partial<PageOptions>) =>
+        dispatch({ type: PageContextActionType.SET_OPTIONS, value: options }),
+      clearOptions: (options: Partial<PageOptions> = initialState.options) =>
+        dispatch({
+          type: PageContextActionType.CLEAR_OPTIONS,
+          value: options,
+        }),
```
ShynRou, line 18: Same as comment above

src/app/containers/PageContainer/PageContainer.tsx
```diff
@@ -0,0 +1,108 @@
...
+        initialOptions.header = HeaderTypes.DEFAULT;
+        initialOptions.footer = FooterTypes.DEFAULT;
+      }
+      actions.clearOptions(initialOptions);
+    },
+    [actions, pageOptions],
+  );
+
+  // reset options to default when route changes
+  useLayoutEffect(() => {
```
ShynRou, line 68: There is no need for this to run after rendering.
```suggestion
  useEffect(() => {
```

src/app/containers/PageContainer/PageContainer.tsx
```diff
@@ -0,0 +1,108 @@
...
+      case HeaderTypes.NONE:
+        return null;
+      case HeaderTypes.DEFAULT:
+        return <DefaultHeaderComponent {...options.headerProps} />;
+      case HeaderTypes.FAST_BTC:
+        return <FastBtcHeader address={options.headerProps?.address} />;
+      case HeaderTypes.CROSS_CHAIN:
+        return <UserWallet address={options.headerProps?.address} />;
+      case HeaderTypes.LABS:
+        return <></>; // todo
```
ShynRou, line 85:
```suggestion
      case HeaderTypes.DEFAULT:
        return <DefaultHeaderComponent {...options.headerProps} />;
      case HeaderTypes.FAST_BTC:
        return <FastBtcHeader address={options.headerProps?.address} />;
      case HeaderTypes.CROSS_CHAIN:
        return <UserWallet address={options.headerProps?.address} />;
      case HeaderTypes.LABS:
        return <></>; // todo
      default:
        return null;
```

src/app/containers/PageContainer/PageContainer.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }, [options.header, options.headerProps]);
+
+  const maybeRenderFooter = useMemo(() => {
+    switch (options.footer) {
+      case FooterTypes.DEFAULT:
+        return <Footer {...options.footerProps} />;
+      default:
+      case FooterTypes.NONE:
+        return null;
```
ShynRou, line 95:
```suggestion
      default:
        return null;
```

ShynRou (changes requested):
Looks already better and I see what you mean with having to set some header props and potentially the header type from within the a page.
Just noted a few code style issues.

JoeStorm99 (dismissed):

Alitasovryn:
FastBTC header

Alitasovryn:
FastBTC header 
Screen Shot 2022-03-18 at 7 47 52 AM

Crosschain header
Screen Shot 2022-03-18 at 8 07 33 AM

FastBTC header has **Back to portfolio** but cross chain header shows **Back.** It would be nice if both header looks similar.

Alitasovryn (dismissed):

ShynRou (approved):

PR #2107, opened 2022-03-23
Feat/2255734392 overlay component

ShynRou:
Resolves https://sovryn.monday.com/boards/2218344956/pulses/2255734392

Testable on /storybook/?path=/story/atoms-overlay--basic
Colors can be testes better when switching the preview background color n the top toolbar.


creed-victor (dismissed):

creed-victor:
Blueprintjs overlay has fade in/out animation, as it's not in spec, I guess we are dropping them?

ShynRou:
@creed-victor yeah, if we really need them, we can reintroduce them easily in this component later.

soulBit (approved):

PR #2116, opened 2022-03-30
Feat: Storybook Design Guide

ShynRou:
Resolves https://sovryn.monday.com/boards/2218344956/pulses/2399958058

Design Guide should be visible on /storybook


ShynRou:
Ahrg... the combined tailwind strings don't work. As they get removed on the ci build… will need to fix that.

ShynRou:
Fixed the tailwind css issues. I had to use the values as style objects since tailwind will just strip out all the dynamic classes.

creed-victor (approved):

creed-victor:
Looks ok, but maybe it would be worth adding example on how to use these classes. For example guideline for border radius, tailwind classname is actually `tw-rounded-{value}`, but from example dev may expect that `tw-radius-{value}` would work

tiltom (dismissed):
Looking good, I just adjusted doc links so they open in a new tab.

creed-victor (dismissed):

tiltom (approved):

creed-victor (approved):

soulBit (approved):

PR #2123, opened 2022-04-06
chore: swap TxDialog with TransactionDialog component

creed-victor:
https://sovryn.monday.com/boards/2218344956/pulses/2394211828


creed-victor:
solved by #2138

PR #2132, opened 2022-04-12
Feat/borrow collateral

creed-victor:
https://sovryn.monday.com/boards/2218344956/pulses/2289263742?userId=26729382


src/app/pages/BorrowPage/components/AddCollateralModal/AddCollateralModal.tsx
```diff
@@ -0,0 +1,336 @@
...
+enum Step {
+  FORM,
+  REVIEW,
+}
+
+export const AddCollateralModal: React.FC<AddCollateralModalProps> = ({
+  loan: item,
+}) => {
+  const canInteract = useCanInteract();
+  const tokenDetails = AssetsDictionary.getByTokenContractAddress(
```
tiltom, line 50: Use `useMemo`.

src/app/pages/BorrowPage/components/AddCollateralModal/AddCollateralModal.tsx
```diff
@@ -0,0 +1,336 @@
...
+    item?.collateralToken || '',
+  );
+  const loanToken = AssetsDictionary.getByTokenContractAddress(
```
tiltom, line 53: Use `useMemo`.

src/app/pages/BorrowPage/components/AddCollateralModal/AddCollateralModal.tsx
```diff
@@ -0,0 +1,336 @@
...
+                    loading={loadingRate}
+                    value={<>{newCollateralRatio} %</>}
+                  />
+                }
+              />
+            </div>
+            <DialogButton
+              confirmLabel={t(translations.common.confirm)}
+              onConfirm={send}
+              disabled={topupLocked || tx.loading || !valid || !canInteract}
```
tiltom, line 318: Extract this condition to a variable and use `useMemo`.

src/app/pages/BorrowPage/components/AddCollateralModal/AddCollateralModal.tsx
```diff
@@ -0,0 +1,336 @@
...
+      calculateCollateralRatio(
+        principalAsCollateral,
+        bignumber(item.collateral).add(weiAmount).toString(),
+      ),
+    [item.collateral, principalAsCollateral, weiAmount],
+  );
+
+  const handleNextStep = useCallback(() => setStep(Step.REVIEW), []);
+
+  return (
```
tiltom, line 122: `render` is way too long, pls extract `Step.FORM` render and `Step.REVIEW` to separate components and use them here, it will help with readability of the code.

src/app/pages/BorrowPage/components/AddCollateralModal/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { bignumber } from 'mathjs';
+
+export const BORROW_MAINTENANCE_RATIO = '115';
```
tiltom, line 3: I am not sure about that. I read the spec (https://sovryn.monday.com/boards/2277102605) and Ororo wrote in comments that users could be already liquidated at 115% and we should use `item.maintenanceMargin` instead (which is hardcoded to 150% currently but I think it's better to be safe than sorry :D ).

src/app/pages/BorrowPage/components/AddCollateralModal/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const calculateLiquidationPrice = (
+  principal: string,
+  collateral: string,
+  ratio: string,
+) =>
+  bignumber(bignumber(principal).mul(bignumber(ratio).div(100)))
+    .div(collateral)
+    .toString();
```
tiltom, line 12:
```suggestion
  bignumber(principal).mul(bignumber(ratio).div(100))
    .div(collateral)
    .toString();
```

src/locales/en/translation.json
```diff
@@ -2187,6 +2189,18 @@
...
+  "addCollateral": {
+    "title": "Add Collateral",
+    "collateralBalance": "Collateral Balance",
+    "liquidationPrice": "Liquidation Price",
+    "collateralRatio": "Collateral Ratio",
+    "amount": "Increase collateral by:",
+    "newInfo": "New details:",
+    "review": "Review Transaction",
+    "errors": {
+      "balance": "Please enter an amount equal to or less than the balance of {{asset}} in your address."
```
tiltom, line 2201:
```suggestion
      "balance": "Please enter an amount equal to or less than your {{asset}} balance."
```

tiltom: I would probably rephrase it this way, it sounds more natural to me than writing "the balance of RBTC in your address" (but I am not a native English speaker so you can consult it with someone more qualified :D ).

tiltom (changes requested):

tiltom:
I don't like the spacing around the error message when you exceed the max amount. I don't think we need to have `margin` there at all since we already have `padding` there + the form group above has `bottom padding` as well.

Screenshot 2022-04-19 at 11.46.55

tiltom:
The processing state of the dialog could use some better title, something like "Adding to collateral" (or at least the capital "P" in "processing").

Screenshot 2022-04-19 at 11 11 41

john-light:
@tiltom QA issues that are not able to left as comments directly on the code should be logged here:
https://sovryn.monday.com/boards/2546493098/views/57703449

It's too easy for stuff to get lost in comments on the PR and it's not possible to reply directly to a comment on a PR so the thread just keeps growing and growing hiding older comments even before they have been resolved.

Can you please move these comments into items on the QA issues board:
https://github.com/DistributedCollective/Sovryn-frontend/pull/2132/#issuecomment-1102415291
https://github.com/DistributedCollective/Sovryn-frontend/pull/2132/#issuecomment-1102411988

creed-victor:
Adding collateral modal and transaction dialog are existing ones in margin trading, so I think changes to the looks should be proposed as new task to the designer to approve and changed globally and not only on this feature

src/app/pages/BorrowPage/components/AddCollateralModal/AddCollateralModal.tsx
```diff
@@ -83,14 +70,19 @@ export const AddCollateralModal: React.FC<AddCollateralModalProps> = ({
     [item.collateral, principalAsCollateral],
   );
 
+  const maintenanceRatio = useMemo(
+    () => bignumber(fromWei(item.maintenanceMargin)).add(100).toFixed(0),
```
tiltom, line 74: I mean, the number is correct but I don't think the way how we got to him is correct as well :D 

I would expect that if you have some percentage value as number somewhere, you should multiply it by 100 and not add 100 to it and I think that our `maintenanceMargin` is incorrect and it should be `1.5` and not `0.15` (because that would give us `15%` instead of `150%` we want). But let me know what you think about it.

image

tiltom (changes requested):

tiltom (approved):

ShynRou (dismissed):

src/app/pages/BorrowPage/components/AddCollateralModal/AddCollateralModal.tsx
```diff
@@ -0,0 +1,169 @@
...
+      calculateCollateralRatio(
+        principalAsCollateral,
+        bignumber(item.collateral).add(weiAmount).toString(),
+      ),
+    [item.collateral, principalAsCollateral, weiAmount],
+  );
+
+  const handleNextStep = useCallback(() => setStep(Step.REVIEW), []);
+
+  const canSubmit = useMemo(
```
tiltom, line 119: This is totally misleading. One would expect that if `canSubmit` is `true` that you actually can submit a tx. But it's the exact opposite and in `CollateralForm` and `ReviewStep` it's used to determine if submit buttons are disabled.

Rename it to something more appropriate (e.g. `isSubmitDisabled`) and also rename `canSubmit` properties on both `CollateralForm` and `ReviewStep` to `disabled`.

src/app/pages/BorrowPage/components/AddCollateralModal/CollateralForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+  tokenDetails: AssetDetails;
+  loanToken: AssetDetails;
+  currentLiquidationPrice: string;
+  currentCollateralRatio: string;
+  loadingRate: boolean;
+  newCollateralAmount: string;
+  newCollateralRatio: string;
+  newLiquidationPrice: string;
+  amount: string;
+  canSubmit: boolean;
```
tiltom, line 32:
```suggestion
  disabled: boolean;
```

src/app/pages/BorrowPage/components/AddCollateralModal/ReviewStep.tsx
```diff
@@ -0,0 +1,100 @@
...
+type ReviewStepProps = {
+  item: ActiveLoan;
+  amount: string;
+  newCollateralAmount: string;
+  newCollateralRatio: string;
+  newLiquidationPrice: string;
+  tokenDetails: AssetDetails;
+  loanToken: AssetDetails;
+  loadingRate: boolean;
+  canSubmit: boolean;
```
tiltom, line 26:
```suggestion
  disabled: boolean;
```

src/app/pages/BorrowPage/components/AddCollateralModal/ReviewStep.tsx
```diff
@@ -0,0 +1,100 @@
...
+      <h1 className="tw-text-sov-white tw-text-center">
+        {t(translations.addCollateral.review)}
+      </h1>
+      <div className="tw-pt-3 tw-pb-2 tw-px-6 tw-bg-gray-2 tw-mb-4 tw-rounded-lg tw-text-sm tw-font-light">
+        <TxFeeCalculator
+          args={[item.loanId, weiAmount]}
+          methodName="depositCollateral"
+          contractName="sovrynProtocol"
+          condition={true}
+          textClassName={'tw-text-gray-10 tw-text-gray-10'}
```
tiltom, line 55:
```suggestion
          textClassName="tw-text-gray-10"
```

src/app/pages/BorrowPage/components/AddCollateralModal/ReviewStep.tsx
```diff
@@ -0,0 +1,100 @@
...
+    <>
+      <h1 className="tw-text-sov-white tw-text-center">
+        {t(translations.addCollateral.review)}
+      </h1>
+      <div className="tw-pt-3 tw-pb-2 tw-px-6 tw-bg-gray-2 tw-mb-4 tw-rounded-lg tw-text-sm tw-font-light">
+        <TxFeeCalculator
+          args={[item.loanId, weiAmount]}
+          methodName="depositCollateral"
+          contractName="sovrynProtocol"
+          condition={true}
```
tiltom, line 54:
```suggestion

```

tiltom: `condition` is an optional property in `TxFeeCalculator` and its default value is `true` so there is no need to pass `true` here.

src/app/pages/BorrowPage/components/AddCollateralModal/ReviewStep.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+type ReviewStepProps = {
+  item: ActiveLoan;
+  amount: string;
+  newCollateralAmount: string;
+  newCollateralRatio: string;
+  newLiquidationPrice: string;
+  tokenDetails: AssetDetails;
+  loanToken: AssetDetails;
+  loadingRate: boolean;
```
tiltom, line 25:
```suggestion
  loading: boolean;
```

src/app/pages/BorrowPage/components/AddCollateralModal/CollateralForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { AssetDetails } from 'utils/models/asset-details';
+import { useWeiAmount } from 'app/hooks/useWeiAmount';
+
+type CollateralFormType = {
+  item: ActiveLoan;
+  tokenDetails: AssetDetails;
+  loanToken: AssetDetails;
+  currentLiquidationPrice: string;
+  currentCollateralRatio: string;
+  loadingRate: boolean;
```
tiltom, line 27:
```suggestion
  loading: boolean;
```

tiltom (changes requested):

JoeStorm99 (approved):

PR #2139, opened 2022-04-14
fix: use timestampToLockDate value instead of current timestamp to calculate voting power

creed-victor:
https://sovryn.monday.com/boards/2218344956/pulses/2539143747

### DIFF
--- a/src/app/hooks/staking/useStaking_timestampToLockDate.ts
+++ b/src/app/hooks/staking/useStaking_timestampToLockDate.ts
@@ -1,0 +1,9 @@
+import { useCacheCallWithValue } from '../useCacheCallWithValue';
+
+export const useStaking_timestampToLockDate = (timestamp: number) =>
+  useCacheCallWithValue(
+    'staking',
+    'timestampToLockDate',
+    '0',
+    Number(timestamp),
+  );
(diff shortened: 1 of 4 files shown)


src/app/hooks/staking/useStaking_timestampToLockDate.ts
```diff
@@ -0,0 +1,9 @@
+import { useCacheCallWithValue } from '../useCacheCallWithValue';
+
+export const useStaking_timestampToLockDate = (timestamp: number) =>
+  useCacheCallWithValue(
+    'staking',
+    'timestampToLockDate',
+    '0',
+    Number(timestamp),
```
tiltom, line 8: Why do you need to cast `timestamp`? It's already a `number`.

tiltom (changes requested):

src/app/hooks/staking/useStaking_timestampToLockDate.ts
creed-victor: Ops

tiltom (approved):

soulBit (approved):

ShynRou (approved):

Alitasovryn (approved):

JoeStorm99 (approved):

PR #2141, opened 2022-04-14
Fix StakePage crashes

ShynRou:
The StakePage crashes, because of two issues introduced by the #1758 merge, this fixes these.


tiltom (approved):

creed-victor (approved):

src/utils/blockchain/math-helpers.ts
```diff
@@ -97,17 +97,10 @@ export const trimZero = (amount: string) => {
...
-    value === null
-  ) {
-    return false;
-  }
-  return true;
-};
+export const isValidNumerishValue = (value: any) =>
+  ['string', 'number'].includes(typeof value) &&
+  value !== '' &&
+  Number.isFinite(Number(value));
```
soulBit, line 103: Not sure if this PR is still needed as staking page appears fine on latest `development` branch, but in case it is - 

this was returning false when `value` is bignumber type (and therefore causing `toValidNumberishValue()` to always return 0), below code adds a check to handle those cases:
```suggestion
export const isValidNumerishValue = (value: any) =>
  (['string', 'number'].includes(typeof value) &&
    value !== '' &&
    Number.isFinite(Number(value))) ||
  (typeof value === 'object' && value.isFinite?.())
```

soulBit (changes requested):

tiltom (approved):

soulBit (approved):

PR #2151, opened 2022-04-21
feat: universal dialog component

creed-victor:
https://sovryn.monday.com/boards/2218344956/pulses/2236624785

Example is in https://feat-refactor-dialog.sovryn.pages.dev/storybook?path=/story/molecule-dialog

Note: had to make close button / icon smaller than it was, otherwise it's going out of viewport on smaller screens (because it's outside of dialog for non-mobile devices)


JoeStorm99 (dismissed):

src/app/components/Dialog/index.module.scss
```diff
@@ -0,0 +1,31 @@
+.dialog_wrapper {
+    @apply tw-fixed tw-inset-0;
+    @apply tw-flex tw-flex-col tw-justify-center tw-items-center;
```
tiltom, line 3:
```suggestion
    @apply tw-fixed tw-inset-0 tw-flex tw-flex-col tw-justify-center tw-items-center;
```

src/app/components/Dialog/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  useMemo,
+  MouseEvent,
+} from 'react';
+import { useTranslation } from 'react-i18next';
+import classNames from 'classnames';
+import { translations } from 'locales/i18n';
+import { Overlay, OverlayProps } from '../Overlay';
+import styles from './index.module.scss';
+
+interface DialogFunctionComponent<T = {}> extends FunctionComponent<T> {
```
tiltom, line 14:
```suggestion
interface IDialogFunctionComponent<T = {}> extends FunctionComponent<T> {
```

tiltom, line 14: Create `types.ts` file and extract this, `DialogSize` and `sizeMap` to it.

src/app/components/Dialog/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onClose,
+}) => {
+  const { t } = useTranslation();
+  const sizeClassNames = useMemo(() => sizeMap[width], [width]);
+
+  const handleClose = useCallback(() => {
+    if (onClose) {
+      onClose();
+    }
+  }, [onClose]);
```
tiltom, line 66:
```suggestion
const handleClose = useCallback(() => {
    onClose?.();
  }, [onClose]);
```

src/app/components/Dialog/index.module.scss
```diff
@@ -0,0 +1,31 @@
+.dialog_wrapper {
```
tiltom, line 1:
```suggestion
.wrapper {
```

src/app/components/Dialog/index.module.scss
```diff
@@ -0,0 +1,31 @@
...
+    @apply tw-fixed tw-inset-0;
+    @apply tw-flex tw-flex-col tw-justify-center tw-items-center;
+}
+
+.dialog_container {
```
tiltom, line 6:
```suggestion
container {
```

src/app/components/Dialog/index.module.scss
```diff
@@ -0,0 +1,31 @@
...
+}
+
+.dialog {
+    @apply tw-relative tw-bg-black tw-text-sov-white tw-px-4 tw-py-8 tw-mx-auto tw-min-h-full;
+    @screen sm {
+       @apply tw-min-h-0 tw-rounded-xl tw-my-16 tw-py-8;
+    }
+}
+
+.close_button {
```
tiltom, line 20:
```suggestion
.closeButton {
```

tiltom (changes requested):

tiltom (dismissed):

src/app/components/Dialog/types.ts
```diff
@@ -0,0 +1,27 @@
+import type { FunctionComponent } from 'react';
+
+export interface IDialogFunctionComponent<T = {}> extends FunctionComponent<T> {
+  index: number;
+}
```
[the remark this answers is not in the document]
creed-victor: Not sure I understood. It's static variable which tracks how many dialogs was opened, so if you would open another dialog while one is already opened - newest would appear on top of the previous one

JoeStorm99 (dismissed):

soulBit (changes requested):
Generally looks good but few things to check before I approve:

1. Point 6a in the spec indicates we should have sizes for `2-7xl` but we only have up to 3 in this PR
2. Point 6c in the spec doesn't seem to be implemented here

@ShynRou if the above two points are fine then let's update the spec to match

I've also made a small fix to the `DialogSize` enum so that the size values appear correctly in Storybook (they were appearing as indexes previously).

creed-victor:
> 1. Point 6a in the spec indicates we should have sizes for `2-7xl` but we only have up to 3 in this PR

Added missing sizes

> 2. Point 6c in the spec doesn't seem to be implemented here

It's implemented but you need to change your storybook preview width to the mobile device (xs size)

tiltom (approved):

soulBit (approved):

JoeStorm99 (approved):

PR #2152, opened 2022-04-21
[RadioButton] - add RadioButton component

pietro-maximoff:
https://sovryn.monday.com/boards/2218344956/pulses/2368699487

### DIFF
--- a/src/app/components/RadioButton/index.tsx
+++ b/src/app/components/RadioButton/index.tsx
@@ -1,0 +1,93 @@
+import React, { useCallback, useMemo, useState } from 'react';
+import classNames from 'classnames';
+import styles from './index.module.scss';
+import {
+  MinimumTwoEntries,
+  RadioButtonColor,
+  RadioButtonSize,
+  RadioButtonStyle,
+} from './types';
+
+type RadioButtonProps = {
+  merged: boolean;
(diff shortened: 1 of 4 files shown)


src/app/components/RadioButton/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    [onChange],
+  );
+
+  const classNameComplete = useMemo(
+    () =>
+      classNames(
+        styles.button,
+        size && styles[size],
+        color && styles[color],
+        style && styles[style],
```
creed-victor, line 50: To me it would be better to have styles mapped for each enum value enforcing typecheck, so it would show error when new size/color/style is added to enums but classes are not added:

```typescript
const colorClassMap: Record<RadioButtonColor, string> = {
   // ...  
  [RadioButtonSize.tradeLong]: 'tw-text-trade-long tw-bg-trade-long tw-border-trade-long',
   // ...
};
```

creed-victor (changes requested):

src/app/components/RadioButton/index.tsx
pietro-maximoff: yes, does make sense

creed-victor (approved):

soulBit (approved):

PR #2156, opened 2022-04-25
Feat: Reduce opacity usages

ShynRou:


ShynRou:
@soulBit this PR breaks the design of some components, and would have required additional work that was included in the Change colors 2 task (can't find it anymore…).
Since there is no further adjustments planned for this DApp, we may want to close it.

soulBit:
Closing this as suggested

PR #2159, opened 2022-04-27
Feat/2255707517 icon atomic component

pietro-maximoff:
https://sovryn.monday.com/boards/2218344956/pulses/2255707517


src/app/components/Icon/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  }, [icon, inline, size]);
+
+  return (
+    <div className={className}>
+      {isFaIcon ? (
+        <FontAwesomeIcon size={iconFaSize} icon={icon as IconProp} />
+      ) : (
+        renderIcon
+      )}
+    </div>
```
ShynRou, line 89: The div wrapper is not needed here, className should be passed directly to FontAwesomeIcon, svg or the customIcon div.
Inline should also set the outer element display to inline-block.

src/app/components/Icon/types.ts
```diff
@@ -0,0 +1,17 @@
...
+import { ReactNode } from 'react';
+import * as IconNames from './iconNames';
+
+export type IconName = typeof IconNames[keyof typeof IconNames];
+
+export enum IconSize {
+  STANDARD = 16,
+  INLINE = '1em',
+  SM = 'sm',
+}
```
ShynRou, line 11: I think these should be constants in the main file and not an enum.

src/app/components/Icon/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * Name of a Sovryn UI icon, or an custom SVG element, or an Fontawesome imported icon to render.
+   * This prop is required because it determines the content of the component
+   */
+  icon: IconType;
+  /**
+   * Size of the icon, in pixels or in scale (1x, 2x, 3x ...) for the Fontawesome icons.
+   * @default IconSize.STANDARD = 16
+   */
+  size?: number | string;
```
ShynRou, line 18:
```suggestion
  size?: number | SizeProp;
```

ShynRou (changes requested):

creed-victor (approved):

soulBit (approved):

tiltom (approved):

PR #2163, opened 2022-05-02
Feat/2426099128 menu enforcement

pietro-maximoff:
https://sovryn.monday.com/boards/2218344956/pulses/2426099128

the main changes in the **header** (desktop and mobile menu), as well as where the button to **connect the wallet** and the **button to change the language**


src/app/containers/PageContainer/components/DefaultHeaderComponent/DefaultHeaderComponent.tsx
```diff
@@ -56,6 +57,18 @@ export const DefaultHeaderComponent: React.FC = () => {
     </button>
   );
 
+  const perpetualPage = {
```
tiltom, line 60: This doesn't work correctly now, it adds an additional separator with nothing in it on mainnet. You need to filter `pages` (see the comment there).

image

src/app/containers/PageContainer/components/DefaultHeaderComponent/DefaultHeaderComponent.tsx
```diff
@@ -152,71 +177,28 @@ export const DefaultHeaderComponent: React.FC = () => {
...
 
-  if (!isMainnet) {
-    pages.push({
-      to: '/perpetuals',
-      title: t(translations.mainMenu.perpetuals),
-      dataActionId: 'header-link-lab-perpetuals',
-    });
-  }
-
   const menuItems = pages.map((item, index) => {
```
tiltom, line 184:
```suggestion
  const menuItems = pages.filter(page => page.title !== '').map((item, index) => {
```

src/app/containers/PageContainer/components/DefaultHeaderComponent/DefaultHeaderComponent.tsx
```diff
@@ -293,269 +282,247 @@ export const DefaultHeaderComponent: React.FC = () => {
...
+                  onClick={() => {
+                    history.push('/wallet');
+                  }}
+                  text={t(translations.mainMenu.wallet)}
+                  data-action-id="header-link-portfolio"
+                  className={`tw-flex tw-flex-row tw-items-center ${
+                    isSectionOpen(SECTION_TYPE.PORTFOLIO) && 'tw-font-bold'
+                  }`}
+                />
+                {currentNetwork === AppMode.TESTNET && (
```
tiltom, line 441: I don't think this is correct - you just don't want to display the link to `perpetuals` on mainnet but you want to display `Labs` section on every env. With this implementation, you don't display the whole `Labs` section on mainnet but you display it in the mobile menu anyway.

tiltom (changes requested):

src/app/containers/PageContainer/components/DefaultHeaderComponent/DefaultHeaderComponent.tsx
pietro-maximoff: yeah, the same problem in the test network and in the main network now, the fixes have been added

tiltom (approved):

creed-victor (approved):

ShynRou (commented):
@soulBit Can you please confirm whether the move of origins into labs is correct?

@pietro-maximoff I think the menu items in the header should have rounded borders, since other buttons are rounded.
image

Otherwise looks good.

Alitasovryn (approved):

PR #2183, opened 2022-05-06
Fix/closing limit order

creed-victor:


creed-victor:
Cherry picked changes #2184

PR #2193, opened 2022-05-09
feat(chart): add graphql for trading chart

creed-victor:
https://sovryn.monday.com/boards/2218344956/views/53321005/pulses/2600564038
https://sovryn.atlassian.net/browse/SOV-356
https://sovryn.atlassian.net/browse/SOV-501


tiltom (changes requested):

src/app/components/TradingChart/helpers.ts
```diff
@@ -0,0 +1,170 @@
...
+export type Bar = {
+  time: number;
+  low: number;
+  high: number;
+  open: number;
+  close: number;
+  volume?: number;
+};
+
+export const candleSticksGQL = (
```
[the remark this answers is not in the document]
creed-victor: Whoops, thought I removed these functions

ShynRou (approved):

tiltom (dismissed):

creed-victor:
Added fixes. Best to test on staging because there is no enough data on testnet

src/app/components/TradingChart/index.tsx
```diff
@@ -89,18 +98,25 @@ export function TradingChart(props: ChartContainerProps) {
...
+  }, [client]);
 
   useLayoutEffect(() => {
     if (chart && hasCharts) {
+      chart.chart().resetData();
+
+      // if quote asset is not RBTC or XUSD, make it line chart, otherwise candle
+      // 1 - candles
+      // 2 - line
+      chart.chart().setChartType(hasDirectFeed(props.symbol) ? 1 : 2);
```
tiltom, line 110: Make this an enum (there is `SeriesStyle` enum in `charting_library` with all of the values).

src/app/components/TradingChart/index.tsx
```diff
@@ -89,18 +98,25 @@ export function TradingChart(props: ChartContainerProps) {
...
-  }, []);
+  }, [client]);
 
   useLayoutEffect(() => {
     if (chart && hasCharts) {
+      chart.chart().resetData();
+
+      // if quote asset is not RBTC or XUSD, make it line chart, otherwise candle
+      // 1 - candles
+      // 2 - line
```
tiltom, line 109:
```suggestion
      // if quote asset is not RBTC or XUSD, make it line chart, otherwise candle
```

tiltom (changes requested):

src/app/components/TradingChart/helpers.ts
```diff
@@ -0,0 +1,273 @@
...
+          totalVolume
+          periodStartUnix
+        }
+      }
+    `;
+  return client
+    .query({
+      query,
+    })
+    .then(response => response.data?.candleSticks || []);
```
soulBit, line 74:
```suggestion
    .then(
      response =>
        response.data?.candleSticks?.map(item => ({
          ...item,
          open: Number(item.open),
          high: Number(item.high),
          low: Number(item.low),
          close: Number(item.close),
          totalVolume: Number(item.totalVolume),
        })) || [],
    )
```

Currently volume does not show up on the chart even when though data is returned from Graph query in `totalVolume` prop. This is due to the `volume` prop passed to the TradingView chart in candle data needing to be type `number` rather than a string. Simple fix is to convert the volume value in `queryCandles()`:
```volume: Number(item.totalVolume.toString()),``` or use suggested code change here.

However, this isn't the ideal solution as the `CandleSticksResponse` type does not have correct types for this prop, or `open`, `low,` `high`, and `close`. These values are never parsed into number format, as the data returned from Graph query is meant to already be in number format as per defined type. Ideally we would have these values returned as numbers directly by Graph, but if this is not possible then we should update typings/logic on FE (e.g. in `fillItem()`).

File: src/app/components/TradingChart/index.tsx [hunk 5]
```diff
@@ -60,6 +67,8 @@ export function TradingChart(props: ChartContainerProps) {
         autosize: true,
         // toolbar_bg: '#a3a3a3',
         theme: props.theme,
+        has_no_volume: false,
```
soulBit, line 70: Let's set this to `true` for any line charts, so that we don't have a non-functional volume indicator showing up. We can re-use the `hasDirectFeed` variable from helpers.tsx

soulBit (changes requested):

src/app/components/TradingChart/index.tsx
creed-victor: Using this type crashes build because of typescript configs, fixing config breaks another part of the dapp. I dont think it's worth debugging just for having typing there

src/app/components/TradingChart/index.tsx
creed-victor: It will re-render entire entire trading view chart when symbol is changed (currently it only shows trading views loading indicator when changing).
Remove it anyway?

soulBit: In that case, it's not worth it - thanks for clarifying

src/app/components/TradingChart/helpers.ts
```diff
@@ -0,0 +1,284 @@
...
+import {
+  CandleDetails,
+  CandleDuration,
+} from 'app/pages/PerpetualPage/hooks/graphql/useGetCandles';
+import { Asset } from 'types';
+import { getTokenContract } from 'utils/blockchain/contract-helpers';
+import { TradingCandleDictionary } from './dictionary';
+
+export const MAX_CANDLE_COUNT = 500;
+export const graph_endpoint: string = process.env.REACT_APP_GRAPH_RSK!;
```
pietro-maximoff, line 11: we have this variable in classifiers (graphRskUrl) and by the way where you use it?

src/app/components/TradingChart/index.tsx
tiltom: Yeah I know you cannot use `SeriesStyle` enum but you can basically copy it to our solution and use it.

creed-victor: Adding, but re-casting to a number will be needed, because type in the setChartType property will not match, even if values are exactly the same.

src/app/components/TradingChart/helpers.ts
creed-victor: you are right, it's not used anymore. Removing it

soulBit (approved):
🎉

Alitasovryn (dismissed):

soulBit:
@creed-victor both testnet/mainnet environment fail to load charts locally, and same on deployed link above or on staging build here:
https://feat-graph-charts--legacy-dapp.netlify.app/spot

creed-victor:
> @creed-victor both testnet/mainnet environment fail to load charts locally, and same on deployed link above or on staging build here: https://feat-graph-charts--legacy-dapp.netlify.app/spot

Where did you find this link? Only this PR only built testnet FE. Testnet should work now, subgraph was protected by CORS

soulBit:
@creed-victor using deployed netlify link (or running locally on testnet) on Spot page:

-  chart data for RBTC/XUSD pair does not load prior to 22nd August (30m candles) - switching timeframes causes more data to load, but stops at differing points for each timeframe. BPRO/DOC pair also has similar issue, so I assume it's a global thing.
- USDT/DOC chart does not load at all

creed-victor:
> @creed-victor using deployed netlify link (or running locally on testnet) on Spot page:
> 
> * chart data for RBTC/XUSD pair does not load prior to 22nd August (30m candles) - switching timeframes causes more data to load, but stops at differing points for each timeframe. BPRO/DOC pair also has similar issue, so I assume it's a global thing.
> * USDT/DOC chart does not load at all

Candles not showing up because there is no data in the subgraph prior 22nd of August.
Will ping Remy to check on the database.

creed-victor:
Ok, so checked code again and it works as intended.
Issue with missing graphs is that there is no trades in some periods and trading view stops requesting for older chunk of data if previous one didn't return data to prevent unnecessary requests.
As XUSD token was changed on testnet and there was not a lot of trades graph is not able to show candles correctly for dates before 22nd of August.
I think testing should be done on staging subgraph instead

tiltom:
> Ok, so checked code again and it works as intended. Issue with missing graphs is that there is no trades in some periods and trading view stops requesting for older chunk of data if previous one didn't return data to prevent unnecessary requests. As XUSD token was changed on testnet and there was not a lot of trades graph is not able to show candles correctly for dates before 22nd of August. I think testing should be done on staging subgraph instead

I tried to test it on staging (locally) but I didn't see any chart on Spot page. It's probably related to the following errors, pls take a look at it @creed-victor .
image

creed-victor:
> I tried to test it on staging (locally) but I didn't see any chart on Spot page. It's probably related to the following errors, pls take a look at it @creed-victor . 
@tiltom, backend on staging / mainnet is not up to date so this is expected

soulBit (approved):

src/app/components/PairNavbarInfo/index.tsx
```diff
@@ -118,13 +120,30 @@ export const PairNavbarInfo: React.FC<IPairNavbarInfoProps> = ({ pair }) => {
     }
   }, [lastPrice, pair, candles, loading, dayPrice]);
 
+  useEffect(() => {
+    const symbol =
+      pair[0].trading_pairs === pair[1].trading_pairs
+        ? `${pair[0].quote_symbol}/${pair[0].base_symbol}`
```
soulBit, line 126: As discussed, Last Traded Price is not populating for some pairs - see comment on Jira for more details

soulBit (changes requested):

soulBit (approved):
🥳

soulBit (approved):

LunaSov (approved):

PR #2223, opened 2022-05-25
Perpetuals limit orders

tiltom:
DO NOT MERGE YET!


src/utils/dictionaries/perpetual-pair-dictionary.ts
```diff
@@ -44,18 +45,20 @@ export class PerpetualPairDictionary {
...
         },
         false,
+        false,
       ),
     ],
     [
       PerpetualPairType.BNBUSD,
       new PerpetualPair(
-        '0xe90b7bceb6e7df5418fb78d8ee546e97c83a08bbccc01a0644d599ccd2a7c2e0',
+        '0xcc69885fda6bcc1a4ace058b4a62bf5e179ea78fd58a1ccd71c22cc9b688792f',
```
creed-victor, line 54: it's some kind of hash in the SC? wit will be the same in the mainnet too?

If it depends on some params, maybe it can be hashed in the FE side: `ethers.utils.id('some-pair-id')`

creed-victor (approved):

soulBit (approved):

src/utils/dictionaries/perpetual-pair-dictionary.ts
tiltom: It's perpetual ID and I am not sure if it stays the same but probably yes.

PR #2246, opened 2022-06-08
chore: change btcs token and aggregator contract

creed-victor:
https://sovryn.monday.com/boards/2218344956/pulses/2780725038


src/utils/blockchain/contracts.testnet.ts
```diff
@@ -324,7 +324,7 @@ export const contracts = {
     chainId: ChainId.BSC_TESTNET,
   },
   PERPETUALS_token: {
-    address: '0xd0370a808203da14B703826eF77072ef5F09840D',
+    address: '0xcF3D22A034Fa157985F0Fe71F15477446f80Be26',
```
tiltom, line 327: I wouldn't adjust it in this PR because even though we can easily test BTC Deposit/Withdraw, we won't be able to trade until we have a new perps deployment that uses the new contract address.

tiltom (changes requested):
I tested both deposit and withdrawal and everything works fine, I'd just leave `PERPETUALS_token` adjustment for the separate PR after new smart contracts deployment.

src/utils/blockchain/contracts.testnet.ts
creed-victor: Understood, maybe you have expected date when perps will have new contracts? If it's soon I think we can leave this change here, and swap it back only if PR will be merged before contracts are upgraded.

tiltom: Once tudor solves bridging problem we have, it shouldn't take too long. So I think it's ok to leave it there,

jjmr007:
According this [acceptance criteria](https://sovryn.monday.com/boards/2218344956/pulses/2780725038?doc_id=2780836369), the condition: 
"User still can deposit BTCS (bsc network) and receive BTC (bitcoin network)" (BTCs -> BTC) is still failing.
Still waiting the payment (in test-BTC; address [tb1qhfq9gmhca8pdvmegcl5eegc6qkrx7pytzrj6qz](https://mempool.space/es/testnet/address/tb1qhfq9gmhca8pdvmegcl5eegc6qkrx7pytzrj6qz)) from this [deposit transaction](https://testnet.bscscan.com/tx/0x8d993580f0f8bcc860f53ef0b67668cbd3007f06838c1e73e08124ea9c7760ab).

PR #2303, opened 2022-06-30
[Lending] - replace legacy backend endpoint usage for Lending APY his…

pietro-maximoff:
https://sovryn.monday.com/boards/2218344956/pulses/2773758352


src/app/hooks/lending/useGetLendingHistory.ts
```diff
@@ -0,0 +1,27 @@
...
+  const [data, setData] = useState<LoanTokenGraphItem[]>([]);
+  const { chainId } = useSelector(selectWalletProvider);
+  const poolAddress = getLendingContract(pool.getAsset()).address;
+
+  useEffect(() => {
+    if (chainId !== undefined) {
+      axios
+        .get(`${subgraphWrapperUrl[chainId]}/lendingApy/${poolAddress}`)
+        .then(res => {
+          setData(res.data.slice(-36)); //last 6 days of data in 4hr chunks
```
tiltom, line 20: `6*4 = 24` if I am not mistaken so either this comment or value doesn't make sense :D Which one of them is it?

src/app/hooks/lending/useGetLendingHistory.ts
```diff
@@ -0,0 +1,27 @@
...
+        })
+        .catch(error => console.log(error));
```
tiltom, line 22: Do we need to `console.log` errors? I'd prefer not to spam the console.

tiltom (changes requested):

BetsyBraddock:
> We show 14-day history on testnet image
> 
> but only 6 days in the PR image
> 
> Is that correct?

I think this is just because the microservice has been up for only 6 days, so we have only 6 days of data. When there is more, it will retrieve 14 days

src/app/hooks/lending/useGetLendingHistory.ts
pietro-maximoff: in one day (24hr) we have 6 points with 4hr interval, 24/4 = 6,
6*6 = 36 items to get last 6 days with 4hr chunks

src/app/hooks/lending/useGetLendingHistory.ts
pietro-maximoff: nope, we don't need it

src/app/hooks/lending/useGetLendingHistory.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function useGetLendingHistory(pool: LendingPool) {
+  const [data, setData] = useState<LoanTokenGraphItem[]>([]);
+  const poolAddress = getLendingContract(pool.getAsset()).address;
+
+  useEffect(() => {
+    if (currentChainId !== undefined) {
+      axios
+        .get(`${subgraphWrapperUrl[currentChainId]}/lendingApy/${poolAddress}`)
+        .then(res => setData(res.data.slice(-36))) //last 6 days of data in 4hr chunks
```
tiltom, line 16:
```suggestion
        .then(res => setData(res.data.slice(-42))) //last 7 days of data in 4hr chunks
```

tiltom: We display one extra day with the current implementation so I think we need to fetch last 7 days.

tiltom (changes requested):

tiltom (approved):
The code itself is fine but we don't have correct data from microservice, see the Betsy's latest comment in Monday.com story . Once this is resolved, it can be merged.

PR #2329, opened 2022-07-13
[Portfolio] - replace endpoint

pietro-maximoff:
https://sovryn.monday.com/boards/2218344956/pulses/2753861718


tiltom:
I see completely different data on live.

New implementation:
image

Existing implementation:
image

pietro-maximoff:
I think it's possible to see different data, @BetsyBraddock can you pls check/confirm it?

tiltom:
I also see different data on testnet.

New implementation:
image

Existing implementation:
image

src/app/containers/VestedHistory/components/VestedPagination.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Button, ButtonStyle } from 'app/components/Button';
+
+interface IVestedPaginationProps {
+  page: number;
+  loading: boolean;
+  isDisabled: boolean;
+  onChange: (value: number) => void;
+}
+
+export const VestedPagination: React.FC<IVestedPaginationProps> = ({
```
tiltom, line 13: I think that the exact same pagination is used for every graph migration story as we don't support proper pagination. I know you cannot reuse it now because neither of the stories is merged but pls create a Monday.com story (in backlog) for creating universal pagination once some of the stories are merged.

src/app/containers/VestedHistory/index.tsx
```diff
@@ -171,37 +65,26 @@ export function VestedHistory() {
...
-                  {t(translations.stake.history.emptyHistory)}
-                </td>
-              </tr>
-            )}
-          {currentHistory && !loading && (
-            <>
-              <VestedHistoryTable items={currentHistory} />
-            </>
+          {isEmpty && (
+            <tr key={'empty'}>
```
tiltom, line 69:
```suggestion
            <tr key="empty">
```

tiltom (changes requested):
I think that you shouldn't delete all of the vesting-related hooks as now you just display Stake history instead of Vesting history. Consult it with @BetsyBraddock pls.

src/app/containers/VestedHistory/components/VestedHistoryRow.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { LinkToExplorer } from 'app/components/LinkToExplorer';
+import { Icon } from 'app/components/Icon';
+import { VestedHistoryFieldsFragment } from 'utils/graphql/rsk/generated';
+import { useDollarValue } from 'app/hooks/useDollarValue';
+import { DisplayDate } from 'app/components/ActiveUserLoanContainer/components/DisplayDate';
+import { getActionName } from '../utils';
+import { AssetValue } from 'app/components/AssetValue';
+import { AssetValueMode } from 'app/components/AssetValue/types';
+
+interface IVestedHistoryRow {
```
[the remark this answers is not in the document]
BetsyBraddock: It looks like here you might be displaying the staking (voluntary staking) history instead of the vesting history. @pietro-maximoff I'll have a look at the code you're using here and will let you know

src/utils/graphql/rsk/generated.tsx
```diff
@@ -8713,6 +8713,35 @@ export type BorrowFieldsFragment = {
...
+    id: string;
+    action: StakeHistoryAction;
+    amount?: string | null;
+    timestamp: number;
+    lockedUntil?: number | null;
+    transaction: { __typename?: 'Transaction'; id: string };
+  }>;
+};
+
+export type VestedHistoryFieldsFragment = {
```
BetsyBraddock, line 8735: @pietro-maximoff  this should be the VestingHistoryItem not the StakeHistoryItem

src/app/containers/VestedHistory/components/VestedPagination.tsx
pietro-maximoff: https://sovryn.monday.com/boards/2218344956/pulses/2968859815

pietro-maximoff:
@BetsyBraddock @tiltom i've added updates in vested history table according to the graphql format of data, check it pls and if its ok for you, then we also need to discuss it with Pili

tiltom:
I don't see my complete history with the new implementation on mainnet. Can you take a look at it @pietro-maximoff and consult it with @BetsyBraddock if needed?

New implementation:
image

Existing implementation:
image

pietro-maximoff:
@tiltom Did clicked on the details button? It should show you the list, like on the margin trading table

tiltom:
> @tiltom Did clicked on the details button? It should show you the sub-table with data, like on the margin trading table

I didn't notice the `Details` button so thanks for the hint but I don't see the complete history anyway (the last item in the new implementation is from 11.1.2022 whereas the last item in the existing implementation is from 18.6.2022).

src/app/containers/VestedHistory/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const getAsset = (type: VestingContractType) => {
+  switch (type) {
+    case VestingContractType.Fish:
+      return AssetsDictionary.get(Asset.FISH);
+    case VestingContractType.FishTeam:
+      return AssetsDictionary.get(Asset.FISH);
+    default:
+      return AssetsDictionary.get(Asset.SOV);
+  }
```
tiltom, line 13:
```suggestion
  switch (type) {
    case VestingContractType.Fish:
    case VestingContractType.FishTeam:
      return AssetsDictionary.get(Asset.FISH);
    default:
      return AssetsDictionary.get(Asset.SOV);
  }
```

tiltom (changes requested):

tiltom (approved):

Alitasovryn (approved):

tiltom (approved):

PR #2336, opened 2022-07-15
feat: allow for claiming of legacy vesting contract SOV fees

soulBit:
https://sovryn.monday.com/boards/2218344956/pulses/2944052071


src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+  ]);
+  const dollars = useCachedAssetPrice(asset, Asset.USDT);
+  const tokenAddress = getContract(token)?.address;
+
+  const currency = useStaking_getLegacyVestingFees(
+    vestingContract,
+    tokenAddress,
+  );
+
+  const dollarValue = useMemo(() => {
```
tiltom, line 46: Use `useDollarValue` (and can get rid of `dollars` as well).

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const { send, ...tx } = useSendToContractAddressTx(
+    vestingContract,
+    VestingABI as AbiItem[],
+    'collectDividends',
+  );
+  const handleWithdrawFee = useCallback(
+    async e => {
+      e.preventDefault();
+      try {
+        send([tokenAddress, 100, account]);
```
tiltom, line 65: Create a const `maxCheckpoints` and use it instead of `100`. Also, just to be sure, use `account.toLowerCase()`.

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+      } catch (e) {
+        console.error(e);
+      }
+    },
+    [tokenAddress, send, account],
+  );
+
+  useEffect(() => updateUsdTotal(contractToken, Number(weiTo4(dollarValue))), [
```
tiltom, line 73: I would probably send as precise numbers as possible and rounded them only when displaying.

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+    contractToken,
+    dollarValue,
+    updateUsdTotal,
+  ]);
+
+  return (
+    <>
+      {
```
tiltom, line 81:
```suggestion

```

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+            text={t(translations.userAssets.actions.withdraw)}
+            disabled={frozen || currency.value === '0'}
+            onClick={handleWithdrawFee}
+            className="tw-lowercase"
+            dataActionId={`staking-withdrawalButton-${asset}`}
+            style={ButtonStyle.link}
+            size={ButtonSize.sm}
+          />
+        </div>
+      }
```
tiltom, line 122:
```suggestion

```

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+    dollarValue,
+    updateUsdTotal,
+  ]);
+
+  return (
+    <>
+      {
+        <div className="tw-flex tw-justify-between tw-items-center tw-mb-1 tw-mt-1 tw-leading-6">
+          <div className="tw-w-2/5">
+            <Tooltip content={<>{t(translations.stake.vestingFeesTooltip)}</>}>
```
tiltom, line 84:
```suggestion
            <Tooltip content={t(translations.stake.vestingFeesTooltip)}>
```

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+              {<>{title || asset} (?)</>}
+            </Tooltip>
+          </div>
+          <div className="tw-w-1/2 tw-mx-4 tw-flex tw-flex-row tw-space-x-2">
+            <div>
+              <Tooltip content={`${weiTo18(currency.value)}`}>
```
tiltom, line 90:
```suggestion
              <Tooltip content={weiTo18(currency.value)}>
```

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+                <LoadableValue
+                  value={weiTo4(currency.value)}
```
tiltom, line 92: Use `AssetValue` component.

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  loading={currency.loading && currency.value === '0'}
+                  loaderContent="0.0000"
+                />
+              </Tooltip>{' '}
+              ≈{' '}
+            </div>
+            <div>
+              <Tooltip content={`${weiToUSD(dollarValue, 6)}`}>
```
tiltom, line 100:
```suggestion
              <Tooltip content={weiToUSD(dollarValue, 6)}>
```

tiltom: I wouldn't round the tooltip value as the whole reason why we show it is to show the most precise value.

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,126 @@
...
+                <LoadableValue
+                  value={weiToUSD(dollarValue)}
```
tiltom, line 102: Instead of using `weiToUSD` method (which is btw really weird), pls use `AssetValue` component.

src/app/containers/StakePage/index.tsx
```diff
@@ -477,14 +483,21 @@ const InnerStakePage: React.FC = () => {
...
-                  title={t(translations.stake.vestingFees)}
-                  useNewContract
-                  frozen={frozen}
-                  vestedFees={true}
-                />
+                {vestingLoading === false &&
+                  vestingItems.length > 0 &&
+                  vestingItems.map(item => {
+                    if (item.type === 'origin') return '';
+                    return (
```
tiltom, line 490:
```suggestion
                    if (item.type === 'origin') {
                       return '';
                    }
                    
                    return (
```

tiltom (changes requested):

src/app/containers/StakePage/index.tsx
```diff
@@ -477,14 +483,21 @@ const InnerStakePage: React.FC = () => {
...
+                      <LegacySovFeeBlock
+                        updateUsdTotal={updateUsdTotal}
+                        contractToken={AssetsDictionary.get(Asset.SOV)}
+                        title={item.type}
+                        frozen={frozen}
+                        vestedFees={true}
```
tiltom, line 496: You don't use `vestedFees` prop in `LegacySovFeeBlock` component at all.

tiltom (changes requested):

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,118 @@
...
+          </div>
+          <div>
+            USD{' '}
+            <AssetValue
+              value={dollarValue.value || '0.0000'}
+              mode={AssetValueMode.auto}
+              useTooltip={true}
+              minDecimals={4}
+              maxDecimals={4}
+            />
```
tiltom, line 102: Instead of doing it manually, you can use `assetString` and `isApproximation` properties in `AssetValue`. If you're ok with having something like `≈ 100 USD` instead of `≈ USD 100`, use it pls.

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
```diff
@@ -0,0 +1,118 @@
...
+  );
+
+  return (
+    <>
+      <div className="tw-flex tw-justify-between tw-items-center tw-mb-1 tw-mt-1 tw-leading-6">
+        <div className="tw-w-2/5">
+          <Tooltip content={t(translations.stake.vestingFeesTooltip)}>
+            {
+              <>
+                {(title && title.charAt(0).toUpperCase() + title.slice(1)) ||
```
tiltom, line 76: This looks like a magic string, pls create a global util function `uppercaseFirstLetter` so it's more clear (if we indeed need to uppercase the first letter).

src/app/containers/StakePage/index.tsx
```diff
@@ -477,14 +483,22 @@ const InnerStakePage: React.FC = () => {
...
-                  contractToken={AssetsDictionary.get(Asset.SOV)}
-                  title={t(translations.stake.vestingFees)}
-                  useNewContract
-                  frozen={frozen}
-                  vestedFees={true}
-                />
+                {vestingLoading === false &&
+                  vestingItems.length > 0 &&
+                  vestingItems.map(item => {
+                    if (item.type === 'origin') {
```
tiltom, line 489: I am not gonna force you to update it but the correct approach would be to change `VestGroup` to an `enum` (and rename it) because this is a magic string. But since it's used in many places and it would just add more complexity to this PR, leave it as is.

tiltom (changes requested):

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
soulBit: Good suggestion, I did already try it that way but it looked awful - will leave it as-is for this PR, but we should add a prop to AssetValue component in Sovryn UI library to allow for placement of the asset symbol before/after the amount

src/app/containers/StakePage/index.tsx
soulBit: Completely agree, I started down this route and realized how much additional refactoring would be required. Marking as resolved

src/app/containers/StakePage/components/LegacySovFeeBlock.tsx
soulBit: We actually already have a helper function for this that I hadn't spotted. Updated to that now

tiltom (dismissed):

creed-victor (approved):

PR #2340, opened 2022-07-15
Update Cookie Consent bar and analytics Opt-out

soulBit:
https://sovryn.monday.com/boards/2218344956/pulses/2946022453

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -11,6 +11,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="Crypto Margin Trading with Sovryn" />
+    <script src="https://cmp.osano.com/16BSnVTBY9ksVax/560923b4-25de-4500-9c44-feefb0dfbf47/osano.js"></script>
     <!-- Google Tag Manager -->
     <script>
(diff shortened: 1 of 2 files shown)


public/index.html
```diff
@@ -11,6 +11,12 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="Crypto Margin Trading with Sovryn" />
+    <style type="text/css">
+      .osano-cm-widget {
+        left: 10px !important;
+      }
+    </style>
+    <script src="https://cmp.osano.com/16BSnVTBY9ksVax/560923b4-25de-4500-9c44-feefb0dfbf47/osano.js"></script>
```
creed-victor, line 19: Be prepared to receive a vulnerability report on immunefi.
There was a discussion some time again about external scripts without integrity checks, I think actual decision is still pending and we had exception for analytics only "for now" because they don't provide it.
If osano allows integrity checks then please add it

soulBit: thanks for the review comment, I've tried to implement this but doesn't appear to be supported from Osano's side:
image

I'll mark this as resolved since we cannot fix this from our end

creed-victor (approved):

PR #2403, opened 2022-08-31
SOV-427: closing limit orders with hw

creed-victor:
Closes https://sovryn.atlassian.net/browse/SOV-427

Two-fold issue:
1. gas price was changing a lot on rsk today causing some transaction not being broadcasted to the nodes
2. chain id was missing on `cancel limit order` tx which prevented HW from signing tx.


src/utils/sovryn/sovryn-network.ts
```diff
@@ -243,18 +231,26 @@ export class SovrynNetwork {
...
+      );
+      setTimeout(() => {
         this.refreshGasPrice();
-      } catch (e) {
-        console.error('gas price update failed.', e);
-        this._store.dispatch(actions.sovrynNetworkError());
-        await this.init(); // Connection aborted, try to reconnect
-      }
-    }, [35e3]); // updates price in 35s intervals (roughly for each block)
+      }, [35e3]); // updates price in 35s intervals (roughly for each block)
```
Rickk137, line 248: Can ```const blockSync = useBlockSync();``` be used here?

src/utils/sovryn/sovryn-network.ts
```diff
@@ -243,18 +231,26 @@ export class SovrynNetwork {
...
+    } catch (e) {
+      console.error('gas price update failed.', e);
```
Rickk137, line 250: what's the correct way of loging? (alternative using``useDebug``)

Rickk137 (changes requested):

src/utils/sovryn/sovryn-network.ts
tiltom: @Rickk137 I don't think that's possible because you can use hooks only in other hooks or components and this file is neither.

src/utils/sovryn/sovryn-network.ts
creed-victor: At this point yes, our logging is hidden by default and user would not see an error and I would like to be able to know that it is this issue if user reports some bug

jjmr007 (dismissed):

soulBit (dismissed):

PR #2426, opened 2022-10-05
hotfix: switch to swap network

creed-victor:


tiltom (commented):
I tested `Swap` and `Spot` and everything works fine for me (except lagging history but it's not a showstopper).

tiltom (approved):

pietro-maximoff (approved):
tested on the margin/spot pages with buy/sell for both - all works well, all transactions are successful and the same on swap page, made 3 random transactions - all works well

LunaSov (approved):

PR #2429, opened 2022-10-07
fix: outdated usage of "allow tokens" contract in bridge

creed-victor:
https://sovryn.atlassian.net/browse/SOV-538

`AllowTokens` contract  for rsk bridge was updated 3 months ago on SC side but not on the FE.
This PR fixes issue by:
1. Changed hardcoded "AllowTokens" address of the bridge to new one
2. Made sure FE checks for "AllowTokens" address from the bridge contract itself and fallbacks to hardcoded value only if request fails.


soulBit (approved):
Nice addition of the fallback 👍🏻

LunaSov (approved):

PR #2436, opened 2022-10-13
Feat/add stmp param for cache/sov 839

BetsyBraddock:
This PR just adds query param ``smtp=Date.now()`` to all calls to the graph-wrapper and amm-apy service to avoid any client-side caching


src/app/containers/SwapFormContainer/index.tsx
```diff
@@ -75,8 +75,16 @@ export const SwapFormContainer: React.FC = () => {
       cancelDataRequest.current = c;
     });
     axios
-      .get(graphWrapperUrl[currentChainId] + 'cmc/summary?extra=true', {
-        cancelToken,
+      .get(graphWrapperUrl[currentChainId] + 'cmc/summary', {
+        params: {
```
soulBit, line 79:
```suggestion
      .get(graphWrapperUrl[currentChainId] + 'cmc/summary', {
        cancelToken,
        params: {
```

cancelToken is unused and the warning is causing build to fail, add it back in

src/app/containers/SwapFormContainer/index.tsx
```diff
@@ -75,8 +75,16 @@ export const SwapFormContainer: React.FC = () => {
...
+          extra: true,
+          stmp: Date.now(),
+        },
+        headers: {
+          'Cache-Control': 'no-cache',
+          Pragma: 'no-cache',
+          Expires: '30',
+        },
```
soulBit, line 87: Add this to getAssetData() call as well, or remove it from here - or is it intentional to only have headers set on this request?

soulBit (changes requested):

PR #2479, opened 2023-02-13
SOV-1457: Add maximum price to order forms

tiltom:
Resolves https://sovryn.atlassian.net/browse/SOV-1457

### DIFF
--- a/src/app/containers/SwapFormContainer/index.tsx
+++ b/src/app/containers/SwapFormContainer/index.tsx
@@ -332,16 +332,16 @@ export const SwapFormContainer: React.FC = () => {
                 <Tooltip
                   content={
                     <AssetValue
-                      value={sourceTokenValue}
-                      assetString={sourceToken}
+                      value={targetTokenValue}
+                      assetString={targetToken}
                       mode={AssetValueMode.auto}
                       maxDecimals={6}
                     />
                   }
                 >
                   <AssetValue
-                    value={targetTokenValue}
-                    assetString={targetToken}
+                    value={sourceTokenValue}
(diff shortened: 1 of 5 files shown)


src/app/containers/SwapFormContainer/index.tsx
```diff
@@ -313,6 +317,18 @@ export const SwapFormContainer: React.FC = () => {
                 <AssetRenderer asset={targetToken} />
               </div>
             </div>
+
+            <div className={styles.swapBtnHelper}>
+              <div>{t(translations.swap.maximumPrice)}</div>
+              <div>
+                <AssetValue
+                  value={maximumPrice}
```
soulBit, line 325: Price is shown as 0 in some cases:
image

And I think price shown should be RBTC/XUSD price (e.g. $20,000) instead of XUSD/RBTC price (1.01 XUSD). 
image

Haven't tested other pairs yet but I assume issue also applies elsewhere. We should already have logic in a hook/util function somewhere to help format this, used for showing entry prices etc, as it's different depending which pair is being displayed.

soulBit (changes requested):

src/app/containers/SwapFormContainer/index.tsx
tiltom: The issue you experienced is based on the fact that `RBTC` and `BNB` on testnet has value of `0 USD` (you can see it on `Portfolio` page as well). The behaviour is not present on mainnet so try to test it there.

I'll take a look at the other issue.

soulBit (dismissed):

soulBit (approved):

PR #2485, opened 2023-02-23
SOV-1865: Added subscription payload to notifications

BetsyBraddock:
https://sovryn.atlassian.net/browse/SOV-1865

### DIFF
--- a/src/app/pages/MarginTradePage/components/NotificationSettingsDialog/index.tsx
+++ b/src/app/pages/MarginTradePage/components/NotificationSettingsDialog/index.tsx
@@ -47,13 +47,11 @@ export const NotificationSettingsDialog: React.FC<INotificationSettingsDialogPro
 
   const emailIsValid = useMemo(() => !email || validateEmail(email), [email]);
   const isSubscribedToD1Notifications = (subscriptions: Subscription[]) => {
-    console.log(notificationUser);
     const subscribed = subscriptions
       .filter(item => item.isSubscribed === true)
       .map(item => item.notification);
     const d1Notifications = Object.values(Notification);
     const output = d1Notifications.every(item => subscribed.includes(item));
-    console.log(output);
     return output;
   };
 
--- a/src/app/pages/MarginTradePage/components/NotificationSettingsDialog/index.tsx
+++ b/src/app/pages/MarginTradePage/components/NotificationSettingsDialog/index.tsx
@@ -23,6 +23,21 @@ interface INotificationSettingsDialogProps {
   onClose: () => void;
 }
 
+const isSubscribedToD1Notifications = (
+  subscriptions: Subscription[] | undefined,
+) => {
+  if (subscriptions === undefined) {
+    return false;
+  } else {
+    const subscribed = subscriptions
+      .filter(item => item.isSubscribed === true)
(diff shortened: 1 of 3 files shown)


src/app/pages/MarginTradePage/components/NotificationSettingsDialog/index.tsx
```diff
@@ -45,6 +46,14 @@ export const NotificationSettingsDialog: React.FC<INotificationSettingsDialogPro
   const [isDiscordActive, setIsDiscordActive] = useState(false);
 
   const emailIsValid = useMemo(() => !email || validateEmail(email), [email]);
+  const isSubscribedToD1Notifications = (subscriptions: Subscription[]) => {
```
soulBit, line 49: You should be able to extract this function to outside of this component definition (so above line 26), currently it redefines new reference to this function every time component renders - not a big thing, but good practice.

src/app/pages/MarginTradePage/components/NotificationSettingsDialog/index.tsx
```diff
@@ -76,6 +85,33 @@ export const NotificationSettingsDialog: React.FC<INotificationSettingsDialogPro
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [notificationUser]);
 
+  const updateSubscriptions = () => {
```
soulBit, line 88: Wrap this function in `useCallback()` with `notificationUser` in dependency array

soulBit (changes requested):

src/app/pages/MarginTradePage/components/NotificationSettingsDialog/index.tsx
BetsyBraddock: I don't think the result of this should be cached - this function is only called when the user actually submits the form. If notificationUser is the dependency array, then when the user submits the form the notificationUser is not updated and the form shows the old version

BetsyBraddock: Thanks soulbit! Making this change now

src/app/pages/MarginTradePage/components/NotificationSettingsDialog/index.tsx
```diff
@@ -76,6 +91,33 @@ export const NotificationSettingsDialog: React.FC<INotificationSettingsDialogPro
...
+    if (notificationUser) {
+      const subscriptions: Subscription[] = JSON.parse(
+        JSON.stringify(notificationUser?.subscriptions),
+      );
+      const d1Notifications = Object.values(Notification);
+      d1Notifications.forEach(item => {
+        const foundSubscription = subscriptions.findIndex(
+          sub => sub.notification === item,
+        );
+        if (foundSubscription) {
```
soulBit, line 104:
```suggestion
        if (foundSubscription >= 0) {
```
Currently crashes if you try to enable notifications with a new wallet as `foundSubscription === -1`, so next line fails

soulBit (changes requested):
Applied changes look good, and functionality works as specified in story. Just one last fix to make

soulBit (approved):

PR #2518, opened 2023-04-06
SOV-2205: Update Reward SOV amounts for D1 pools

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-2205

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -2720,7 +2720,7 @@
       "text": "Provide any amount of XUSD to the XUSD lending pool and instantly start accruing your share of 2,500 SOV rewards."
     },
     "card3": {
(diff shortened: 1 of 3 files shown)


creed-victor (dismissed):

tiltom (changes requested):
I agree with @creed-victor, you should pass localized value to translations from components. If that's too difficult, change it to `7.5K` at least.

pietro-maximoff:
> I agree with @creed-victor, you should pass localized value to translations from components. If that's too difficult, change it to `7.5K` at least.

@tiltom I also agree with that but it needs more time do it, so will change just `,` to `.`

tiltom (dismissed):

PR #2541, opened 2023-05-24
SOV-2301: Update FeeSharing withdraw method on Reward page

Rickk137:
https://sovryn.atlassian.net/browse/SOV-2301

### DIFF
--- a/src/app/pages/RewardPage/components/ClaimForms/FeesEarnedClaimRow/index.tsx
+++ b/src/app/pages/RewardPage/components/ClaimForms/FeesEarnedClaimRow/index.tsx
@@ -39,6 +39,8 @@ interface IFeesEarnedClaimRowProps extends IClaimFormProps {
   assetClaimLocked?: boolean;
 }
 
+const MAX_PROCESSIBLE_CHECKPOINTS = 300;
+
 export const FeesEarnedClaimRow: React.FC<IFeesEarnedClaimRowProps> = ({
   amountToClaim,
   contractAddress,
@@ -105,7 +107,10 @@ export const FeesEarnedClaimRow: React.FC<IFeesEarnedClaimRowProps> = ({
   );
 
   const maxWithdrawCheckpoint = useMemo(
-    () => (Number(maxCheckpoints) > 300 ? '300' : maxCheckpoints),
+    () =>
+      Number(maxCheckpoints) > MAX_PROCESSIBLE_CHECKPOINTS
+        ? String(MAX_PROCESSIBLE_CHECKPOINTS)
+        : maxCheckpoints,
     [maxCheckpoints],
   );
   const onSubmit = useCallback(() => {
--- a/src/app/pages/RewardPage/hooks/useGetNextPositiveCheckpoint.ts
+++ b/src/app/pages/RewardPage/hooks/useGetNextPositiveCheckpoint.ts
@@ -20,7 +20,10 @@ export const useGetNextPositiveCheckpoint = (
   const updateNextPositiveCheckpoint = useCallback(async () => {
     let index = 1;
 
-    while (totalTokenCheckpoints >= MAX_NEXT_POSITIVE_CHECKPOINT * index) {
+    while (
+      totalTokenCheckpoints >=
+      MAX_NEXT_POSITIVE_CHECKPOINT * (index - 1)
+    ) {
       const {
         hasFees,
         checkpointNum,
@@ -38,9 +41,9 @@ export const useGetNextPositiveCheckpoint = (
 
       if (!!hasFees) {
         return setUserCheckpoint({
-          checkpointNum: checkpointNum,
-          hasFees: hasFees,
-          hasSkippedCheckpoints: hasSkippedCheckpoints,
+          checkpointNum,
+          hasFees,
+          hasSkippedCheckpoints,
         });
       }
       index++;


src/app/pages/RewardPage/components/ClaimForms/FeesEarnedClaimRow/index.tsx
```diff
@@ -55,42 +56,100 @@ export const FeesEarnedClaimRow: React.FC<IFeesEarnedClaimRowProps> = ({
...
-    if (asset === Asset.RBTC) {
-      withdrawRBTC(
-        [0, address],
-        { from: address, gas: gasLimit[TxType.STAKING_REWARDS_CLAIM_RBTC] },
-        { type: TxType.STAKING_REWARDS_CLAIM_RBTC },
-      );
+    if (userCheckpoint?.hasSkippedCheckpoints) {
+      if (asset === Asset.RBTC) {
+        withdrawRBTCStartingFromCheckpoint(
+          [userCheckpoint?.checkpointNum, maxCheckpoints, address],
```
tiltom, line 111: I think that you can pass `maxCheckpoints` to all the withdrawing methods only if it's less than `300` (according to https://sovryn.atlassian.net/browse/SOV-2301?focusedCommentId=16960). I suggest creating a variable with `useMemo` for that.

src/app/pages/RewardPage/components/ClaimForms/FeesEarnedClaimRow/index.tsx
```diff
@@ -55,42 +56,100 @@ export const FeesEarnedClaimRow: React.FC<IFeesEarnedClaimRowProps> = ({
...
+      }
     } else {
-      send(
-        [contractAddress, maxCheckpoints, address],
-        { from: address, gas: gasLimit[TxType.STAKING_REWARDS_CLAIM] },
-        { type: TxType.STAKING_REWARDS_CLAIM },
-      );
+      if (asset === Asset.RBTC) {
+        withdrawRBTC(
+          [0, address],
```
tiltom, line 130: Why do you use `0` here if you're using `maxCheckpoints` in `withdraw`?

src/app/pages/RewardPage/hooks/useGetNextPositiveCheckpoint.ts
```diff
@@ -0,0 +1,62 @@
...
+  contractAddress: string,
+  totalTokenCheckpoints: number,
+) => {
+  const address = useAccount();
+  const [userCheckpoint, setUserCheckpoint] = useState<UserCheckpoint>();
+
+  const updateNextPositiveCheckpoint = useCallback(async () => {
+    let index = 1;
+
+    while (index < 5) {
```
tiltom, line 23: Why 5?

src/app/pages/RewardPage/hooks/useGetNextPositiveCheckpoint.ts
```diff
@@ -0,0 +1,62 @@
...
+        });
+      }
+      index++;
+    }
+
+    setUserCheckpoint({
+      checkpointNum: '0',
+      hasFees: false,
+      hasSkippedCheckpoints: false,
+    });
```
tiltom, line 52: I think you can just initialize it on line `18` or is there some reason why you want to do it here?

src/app/pages/RewardPage/hooks/useGetNextPositiveCheckpoint.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      if (
+        !!checkpoint.hasFees ||
+        totalTokenCheckpoints < MAX_NEXT_POSITIVE_CHECKPOINT * index
+      ) {
+        return setUserCheckpoint({
+          checkpointNum: checkpoint.checkpointNum,
+          hasFees: checkpoint.hasFees,
+          hasSkippedCheckpoints: checkpoint.hasSkippedCheckpoints,
+        });
```
tiltom, line 43:
```suggestion
     return setUserCheckpoint({ ...checkpoint });
```

tiltom (changes requested):

src/app/pages/RewardPage/components/ClaimForms/FeesEarnedClaimRow/index.tsx
tiltom: I don't know that so verify it with @tjcloa.

src/app/pages/RewardPage/hooks/useGetNextPositiveCheckpoint.ts
tiltom: I don't think that's true either because he only says that `withdraw` / `withdrawRBTC` methods can process up to 300 checkpoints in one tx. Verify it with @tjcloa pls.

src/app/pages/RewardPage/hooks/useGetNextPositiveCheckpoint.ts
tiltom: That would be fine, the props you need are there and you don't mind extra props. But if you're worried about that, you can always do this:
`const { checkpointNum, hasFees, hasSkippedCheckpoints } = checkpoint;

        return setUserCheckpoint({
          checkpointNum,
          hasFees,
          hasSkippedCheckpoints,
        });
        `

src/app/pages/RewardPage/components/ClaimForms/FeesEarnedClaimRow/index.tsx
creed-victor: To my understanding it should be capped to the 300 per transaction. So if maxCheckpoints > 300, you should use 300. Otherwise tx would fail as it would likely use more gas than block can handle. User will have to make second transaction to withdraw remaining checkpoints afterwards.

src/app/pages/RewardPage/hooks/useGetNextPositiveCheckpoint.ts
tjcloa: max params for getNextPositiveCheckpoint and for withdraw are independent  
`getNextPositiveCheckpoint`: 75 is empirical max value before a node makes the func call fail, so it can digest 75 checkpoints at most per call. we tested it on a local forked mainnet, in theory can be a different number for the actual mainnet. therefore to find the first "claimable" checkpoint with hasFees === true, you need to cycle it in chunks of 75 till you find a claimable checkpoint or process all the checkpoints.
`withdraw functions`: 300 checkpoints at max per call on a local forked mainnet which should be the same as on the mainnet, but could be different in the wild - this is applied when you do actual claiming (withdraw)

creed-victor (approved):

PR #2566, opened 2023-07-26
SOV-2785: Fix max withdrawal logic in D1 Lend page

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-2785

### DIFF
--- a/src/app/pages/LendingPage/components/LendingDialog/index.tsx
+++ b/src/app/pages/LendingPage/components/LendingDialog/index.tsx
@@ -128,6 +128,11 @@ export const LendingDialog: React.FC<ILendingDialogProps> = ({
     }
   }, [depositedAssetBalance, marketLiquidity]);
 
+  const validateRedeem = useMemo(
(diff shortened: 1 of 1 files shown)


src/app/pages/LendingPage/components/LendingDialog/index.tsx
```diff
@@ -94,15 +101,33 @@ export const LendingDialog: React.FC<ILendingDialogProps> = ({
...
+    weiAmount,
+    marketLiquidity,
+  ]);
 
   const [withdrawAmount, setWithdrawAmount] = useState(
     calculateWithdrawAmount(),
   );
 
+  const maxWithdrawAmount = useMemo(() => {
+    if (bignumber(depositedAssetBalance).greaterThan(marketLiquidity)) {
```
soulBit, line 124: Same logic should be applied to Withdraw button disabled state on line 92 in `validate` variable (currently the user is prevented from pushing through a tx which will fail, but the button is still enabled which is confusing)

soulBit (changes requested):

soulBit (approved):

